PR #35, opened 2019-05-31
Fix "example hardware configurations"

kkallday:
It was 404-ing so I linked to the coreos docs.


nate-double-u:
Thanks for this @kkallday, but as we've been making some fairly big changes, I'm closing this PR in favour of https://github.com/etcd-io/website/pull/174

Issue #44, opened 2019-09-01
The docs in the website are all leading to etcd version  2, links to version 3 are broken


oz123:
When I go to the documentation of etcd3 I expect to find the most updated docs, older versions should be easy to find.
Currently, the default is documentation of etcd2 

![image](https://user-images.githubusercontent.com/1083045/64076085-762da200-ccc0-11e9-830e-5bdb0dc70509.png)

The links to updated documentation are broken:

![image](https://user-images.githubusercontent.com/1083045/64076109-a8d79a80-ccc0-11e9-987e-6928e92129ec.png)

PR #47, opened 2019-09-19
content/docs: remove v2

gyuho:
Fix https://github.com/etcd-io/website/issues/44.

Other versions do not have v2 docs.


philips:
Makes sense.

Is there automation that caused this to be copied that needs to be updated as well?

gyuho:
For v3.4, I just manually copied the directory. Followed https://github.com/etcd-io/website#releasing-a-new-version-of-the-etcd-documentation.

PR #78, opened 2020-10-23
Added a missing "/" for go etcd client url

prabhakhar:
Added a missing "/" for go etcd client url. Currently it results in 404


nate-double-u (approved):
Good catch, thanks @prabhakhar!

PR #79, opened 2020-11-05
Fix broken fio-blog-post link in op-guide/hardware

rissson:


nate-double-u (approved):
Thanks for this @rissson!

Issue #74, opened 2020-08-15
Docs for AWS say "across at least two AZs", shouldn't that be three?


cakoose:
In the "Availability" section of https://etcd.io/docs/v3.4.0/platforms/aws/

> In the context of an etcd cluster this means a cluster must contain a minimum of three members where EC2 instances are spread across at least two AZs in order for an etcd cluster to be considered highly available at a Regional level.

If we spread our three etcd nodes across two AZs, then one AZ will have two nodes.  If that AZ goes down, etcd will be completely unavailable, right?

Am I misunderstanding the terms here?

geoffcline:
👀 I'll take a look

patrocinio:
Hello @cakoose You are right: it should say at least 3 AZs.

patrocinio:
Hey @cakoose I see the Platform pages have been removed from recent documentation version (v3.5 and v3.6), and that v3.4 is not being maintained, so it seems this issue is outdated.

Thanks

PR #84, opened 2020-12-08
clarify AWS availability zones and

geoffcline:
in response to issue #74 

asked @gyuho to confirm this guidance is still correct


cakoose:
Hi!  I was the person who filed #74.

Let's say I have 1 node in one AZ an 2 nodes in another AZ.  According to the guide, this is enough to be "highly available within an AWS region".  But if the AZ with 2 nodes goes down, the etcd cluster is unavailable, right?

I'm interpreting "highly available within an AWS region" as "can handle a single AZ failure".  Does it actually means something else?

PR #85, opened 2020-12-09
Distinguish between V2 and V3 in the Etcd tools doc

m0un10:
Having spent my time down this rabbit hole, I wanted to help others by removing repos that no longer exist and making it clear which tools only work on V2 or have other noteworthy constraints.

### DIFF
--- a/content/docs/v3.2.17/integrations.md
+++ b/content/docs/v3.2.17/integrations.md
@@ -4,19 +4,23 @@ title: Libraries and tools
 
 **Tools**
 
-- [etcdctl](https://github.com/coreos/etcd/tree/master/etcdctl) - A command line client for etcd
-- [etcd-backup](https://github.com/fanhattan/etcd-backup) - A powerful command line utility for dumping/restoring etcd - Supports v2
-- [etcd-dump](https://npmjs.org/package/etcd-dump) - Command line utility for dumping/restoring etcd.
-- [etcd-fs](https://github.com/xetorthio/etcd-fs) - FUSE filesystem for etcd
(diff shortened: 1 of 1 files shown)


nate-double-u:
Hi @m0un10, thanks for this. I'm hesitant to approve this PR as it's affecting such an old version of the docs -- I'd tend to prefer to leave the documentation for anything before our _current_ and _latest_ (v3.4.0) versions as a record.

Do you have any thoughts on this @spzala?

m0un10:
@nate-double-u, I thought i was updating etcd's main docs. That page was where google sent me.
Can you please share the link to the latest docs website?

nate-double-u:
> @nate-double-u, I thought i was updating etcd's main docs. That page was where google sent me.
> Can you please share the link to the latest docs website?

Same site, same repo, it's just the folder:
`content/docs/current/` and `content/docs/v3.4.0/` are the currently maintained versions.

m0un10:
When I raised this PR, I suspect that the page I updated was the current version. If it is updated to be against the relevant page that represents the current version now, would you consider accepting this PR? There are still integrations listed that are broken links or V2 only even in the current doc site as far as I can tell.

nate-double-u:
We should probably fix broken links where we find them (though, that's probably best done through opening new issues/PRs). 
Regarding reorganizing the page, I think I'd defer to @spzala.

m0un10:
The driver is not to reorganise but to avoid misleading the user that there
is an integration for their Etcd V3 instance when in fact it only works
against V2. Another way could be to just remove the V2 only integrations
from V3 docs in addition to removing the broken links. Would you be to open
to this?

PR #91, opened 2021-01-13
Fix faulty Go libraries links.

atompi:
Fix faulty Go libraries links.

etcd/v2

etcd/v3


nate-double-u:
Hi @atompi, thanks for this. Would you be willing to fix the conflicts, and update the `integrations.md` file in `current` as it likely has the same  issue?

atompi:
> Hi @atompi, thanks for this. Would you be willing to fix the conflicts, and update the `integrations.md` file in `current` as it likely has the same issue?

Of course, I've fixed it.

nate-double-u (approved):

PR #92, opened 2021-01-16
Update integrations.md

chris-makaio:
remove etcdsh (repo doesn't exist  anymore)


nate-double-u:
Thanks @chris-makaio, would you be willing to make this change to the `current` folder `integrations.md` file as well, I expect it has the same issue.

PR #100, opened 2021-01-28
Update link to zookeeper docs

nemobis:
The point is still valid I guess, although there were some developments.


nate-double-u:
Thanks for this @nemobis, could we make this change to the `current` `/learning/why.md` file instead of the v3.4.0 file? Probably best to let v3.4.0 be a historical record at this point.

nate-double-u:
Hi @nemobis, with all the folder changes made around issue https://github.com/etcd-io/website/issues/152, probably the only folders needing change are now `next` and `v3.4`

nate-double-u:
As there have been a lot of changes to the site, I'm closing this in favour of #203.

PR #102, opened 2021-02-09
Fix code snippet

spzala:
Noticed that the import path is incorrect and clus is not declared, whereas, cli is unused. Updating
snippet to make it working.


ptabor (approved):

spzala:
Thanks for the quick review @ptabor

Issue #90, opened 2021-01-13
Broken links 


spzala:
While switching to a different version in some cases, the resulting link gives 404 error. 
For example, try to open any version (e.g. v3.4.0) from https://etcd.io/docs/v2/backward_compatibility/ The resulting link https://etcd.io/docs/v3.4.0/backward_compatibility/ is broken. 

/cc @nate-double-u

nate-double-u:
I think this may be due to some pages that have been removed between v2 and v3.

I wonder if there is a better way of handling this 404 in these cases.

We may also want to look into some sort of link checking system generally to try to automate catching these.

nate-double-u:
Netlify looks like it supports 404s based on path:
https://docs.netlify.com/routing/redirects/redirect-options/#custom-404-page-handling

nate-double-u:
Related issue: https://github.com/etcd-io/website/issues/87

nate-double-u:
> I think this may be due to some pages being removed, or moved, between v2 and v3.

We could put together a UX solution to this -- right now we have a menu bar of all the versions, I'm thinking that if there isn't a page available in a given version that version's box gets greyed out. If clicked, it would bring you to that version's root page. The drop down menu should probably also bring the reader to that version's root page (rather than the 404).

Issue #98, opened 2021-01-27
Migrate documentation: Site cleanup / prep


nate-double-u:
## Description

As per the conversation around the [Etcd.io Docs/SEO Improvement Plan](https://github.com/etcd-io/website/issues/65):

Cleanup the existing site architecture (Bring prose out of config files, layout files, etc., and make the site generally more hugo-like)

## Deliverable

Site prepared for Docsy conversion

## Notes

- https://docs.google.com/spreadsheets/d/1-ZQMPc_eQ0fh1pwOHv3NltwT-ifKA5UcFePfdFiWn8I/edit?usp=sharing

nate-double-u:
Closing as we're at the point of creating a docsy branch and moving on to that process. Any additional cleanup will be apart of https://github.com/etcd-io/website/issues/94 or https://github.com/etcd-io/website/issues/126.

PR #108, opened 2021-02-19
Cleanup/prep

nate-double-u:
- Updating Makefile
  - Updating DOCKER_IMG, removing unneeded lines
- removing data/faq.yaml and adding any missing questions to content/docs/current/faq.md
- removing "/docs /docs/{{ $latest }}" redirect as it appears not to be working, and we may want to use /docs in future as part of the docs TOC
- internal links updates
- adding list of generated metrics to metrics page
  - adding table of generated metrics
- updating /docs/current/metrics
- removing etcd-mixins from current as they are not Documentation and have been moved in the etcd-io/etcd repo.
- general links cleanup
- Updating versions system (drop-downs, "Versions of this doc",
deprecation warnings)
  - Versions links now default to a versions top level if there is no file associated with the current version in the linked to version

contributes to: https://github.com/etcd-io/website/issues/98
fixes: https://github.com/etcd-io/website/issues/90


nate-double-u:
/cc @spzala @ptabor 

I've gone through and think that all the links in `content/docs/current` work -- with the exception of files that are referenced in the version menu that are not in other versions. I'm still not sure how to deal with the versioning system on that issue.

I'm also not sure about what to do about the following:
1. What should we do with the `content/docs/current/learning/lock` folder & content? I've removed it as it no longer makes sense as a README, and it was providing an example of how to do something as if the reader were still grabbing the info from the code repo. Instead of removing it, should we rework it to make more sense in the `learning` section?
2. Should we remove `content/docs/current/dev-guide/api-spec/swagger`? Is it used anywhere else? Is it documentation?

nate-double-u:
Deploy preview:
https://deploy-preview-108--etcd.netlify.app
https://deploy-preview-108--etcd.netlify.app/docs/current/metrics/#generated-list-of-metrics

nate-double-u:
I haven't moved any copy out of the config files yet, looking at how the site is built, and how the gRPC docsy conversion went, I think that the moving of the copy is a part of the work of doing that conversion.

nate-double-u:
In my last force-push, I've redone the versioning system -- now, if a generated link in a drop-down, deprecation warning, or in the "Versions of this doc" list don't have a corresponding file in the target version, the link will default to the top level of the target version instead.

This has reduced our broken links reports from 1651 to 167. I'm still looking at the remaining broken links, but i think this fixes: https://github.com/etcd-io/website/issues/90

nate-double-u:
Versioning fixes deploy previews:

Here are some pages that demonstrate how the new versioning system works (drop down menu, 'versions of this doc' line, and deprecation warning)

Every version has a FAQ page:
https://deploy-preview-108--etcd.netlify.app/docs/v2/faq/

Only version v2 and v3.1.12 have a Libraries and Tools page:
https://deploy-preview-108--etcd.netlify.app/docs/v2/libraries-and-tools/

Run etcd on Containel Linux with systemd was introduced in version v3.2.17:
https://deploy-preview-108--etcd.netlify.app/docs/v3.2.17/platforms/container-linux-systemd/

spzala:
> /cc @spzala @ptabor
> 
> I've gone through and think that all the links in `content/docs/current` work -- with the exception of files that are referenced in the version menu that are not in other versions. I'm still not sure how to deal with the versioning system on that issue.
> 
> I'm also not sure about what to do about the following:
> 
>     1. What should we do with the `content/docs/current/learning/lock` folder & content? I've removed it as it no longer makes sense as a README, and it was providing an example of how to do something as if the reader were still grabbing the info from the code repo. Instead of removing it, should we rework it to make more sense in the `learning` section?
> 
>     2. Should we remove `content/docs/current/dev-guide/api-spec/swagger`? Is it used anywhere else? Is it documentation?

@nate-double-u yes, I would suggest to keep both for now and we can rethink it as a separate issue? Thanks!

spzala (commented):
Where are putting deleted learning/lock files? I am sure you are taking care of it, but got little lost :) Thanks @nate-double-u 
Otherwise looks great.

nate-double-u:
I'll put them back now -- I had taken them out when I made the comment. Next push will restore them

spzala:
> I'll put them back now -- I had taken them out when I made the comment. Next push will restore them

OK, cool. Thanks!

nate-double-u:
I'm marking this PR as "ready for review" as I'm not a fan of really big PRs, and this one is getting to be quite large. If we need to do more cleanup, we can do so with a new issue/PR.

spzala (approved):

spzala:
> I'm marking this PR as "ready for review" as I'm not a fan of really big PRs, and this one is getting to be quite large. If we need to do more cleanup, we can do so with a new issue/PR.

Yup, sounds great. Thanks @nate-double-u

nate-double-u:
Thanks Sahdev!

PR #117, opened 2021-03-16
Gitignore: ignore public even if it's a symlink

chalin:
This tweak allows us to have a symlink for `public`, which is useful when contributors use another repo to hold the generated site files for the purpose of identifying site changes.

cc @spzala @nate-double-u

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 node_modules/
 
 # Hugo-generated assets
-public/
+/public
 resources/
 
 # Link checker artifacts


.gitignore
```diff
@@ -2,7 +2,7 @@
 node_modules/
 
 # Hugo-generated assets
-public/
+/public
```
nate-double-u, line 5: I've been using gitignore.io for a while, and they suggest `/public/` 
(https://www.toptal.com/developers/gitignore/api/hugo -- or an editable version: https://www.toptal.com/developers/gitignore?templates=hugo).

While we are in the .gitignore, should we flush it out?

chalin: > Would this preserve the intent of this PR?

No. As part of my workflow (and one I suggest using when doing significant rework to a site), I symlink `/public` to `../website.g` -- a local git repo used to hold the generated site files. Diffing the generated site files is very useful to ensure that a site isn't change after a refactoring, to to measure the impact of a shortcode change, etc.

Having a trailing slash prevents a `public` symlink from being ignored. Removing the trailing slash is harmless (in our case), otherwise.

spzala (approved):
lgtm
Thanks @chalin @nate-double-u

PR #124, opened 2021-03-19
Generate production-mode CSS in production mode only

chalin:
Hugo docs don't use the term "mode", they instead use "environment", but that just makes things confusing. For details concerning the development vs. production "mode" (i.e., "environment"), see [Hugo > Functions > hugo](https://gohugo.io/functions/hugo/).

(This file needs some more cleanup, but I wanted to keep diffs to a minimum for now.)


chalin:
Note that this change doesn't change the production generated site.

nate-double-u (approved):

Issue #152, opened 2021-03-24
Revise Versioning Structure to MAJOR.MINOR


nate-double-u:
During the [March 11, 2021 etcd public meeting ](https://docs.google.com/document/d/16XEGyPBisZvmmoIHSZzv__LoyOeluC5a4x353CX0SIM/edit#heading=h.2pcfcuck4vc) the question of when to cut new docs releases came up, and consensus seemed to be on MAJOR.MINOR releases of the main code.

I suggest that we structure the documentation releases the same way. This would affect both the underlying folder structure, and the version displayed on the site.

| Existing version | Proposed version | Update completed |
| --- | --- | --- |
| `current` | `next` | [ ] https://github.com/etcd-io/website/pull/161 |
| `v3.4.0` | `v3.4` | [ ] https://github.com/etcd-io/website/pull/161 |
| `v3.3.13` | `v3.3` | ✅ https://github.com/etcd-io/website/pull/159 |
| `v3.3.12` | removed | ✅ https://github.com/etcd-io/website/pull/159 |
| `v3.2.17` | `v3.2` | ✅ https://github.com/etcd-io/website/pull/158 |
| `v3.1.12` | `v3.1` | ✅ https://github.com/etcd-io/website/pull/156 |
| `v2` | `v2.3` | ✅ https://github.com/etcd-io/website/pull/154 |

* _Note_: I'm not sure what the most recent version of `v2` is, will need to look it up before making the change. [edit] Looks like the last v2 release was [v2.3.7](https://github.com/etcd-io/etcd/releases/tag/v2.3.7). I've updated the table to reflect this.

Part of the rationale for this is to simplify the version selector code, as well as to bring the documentation in line with what the actual release cadence will be moving forward.

nate-double-u:
@spzala @jberkus @gyuho @jingyih @ptabor @xiang90 @jpbetz @hexfusion @chalin

ptabor:
That's LGTM. Thank you.

hexfusion:
+1

jpbetz:
+1

gyuho:
+1

nate-double-u:
Added a column to the table to track the work and PR numbers.

nate-double-u:
`v3.3.13` ➔ `v3.3` and the removal of `v3.3.12` should be done at the same time. We can set the redirect for v3.3.12 to point at the new `v3.3`.

nate-double-u:
[...]

PR #154, opened 2021-03-24
Revise Versioning Structure v2 --> v2.3

nate-double-u:
Moving files. 

Updating config.toml, partials/docs/hero.html, partials/docs/nav-panel.html, and content/docs/v2.3/_index.md

Contributes to: https://github.com/etcd-io/website/issues/152

/cc @chalin


nate-double-u:
deploy preview: https://deploy-preview-154--etcd.netlify.app/

chalin:
And the `_redirects`?

nate-double-u:
> And the `_redirects`?

Right!

nate-double-u:
@chalin, I've added the `static/_redirects` file, and the redirect rules check is coming back passed, but the deploy preview doesn't seem to actually redirect. 

https://deploy-preview-154--etcd.netlify.app/docs/v2/ <- should show https://deploy-preview-154--etcd.netlify.app/docs/v2.3/

Thoughts?

nate-double-u:
actually, the redirect check is still only checking 4 rules, so netlify isn't seeing `static/_redirects`

chalin:
I know that it isn't obvious, but the file to change is https://github.com/etcd-io/website/blob/master/layouts/index.redirects. Sorry for not being clearer earlier, but your name shows up in the changes to that file so I thought that you knew / remembered.

chalin:
(It's the _generated_ file that is named `_redirects`.)

nate-double-u:
I've tried updating `index.redirects` as follows:

```
/docs/latest     /docs/v3.4.0
/docs/latest/*     /docs/v3.4.0/:splat
/docs/v2/*    /docs/v2.3/:splat 301!
```
which generates a `_redirects` file of: 
```

/docs/latest     /docs/v3.4.0
/docs/latest/*     /docs/v3.4.0/:splat
/docs/v2/*    /docs/v2.3/:splat 301!
```
But this doesn't actually forward the url (at least with `netlify dev`)

Updating `netlfy.toml` with:
```
[[redirects]]
from = "/docs/v2/*"
to = "/docs/v2.3/:splat"
status = 301
force = true
```
only works (with `netlify dev` if you use a trailing slash).

netlify.toml
```diff
@@ -33,3 +33,8 @@ to = "/docs/v3.4.0/reporting_bugs"
 status = 301
 force = false
 
+[[redirects]]
+from = "/docs/v2/*"
+to = "/docs/v2.3/:splat"
+status = 301
+force = true
```
chalin, line 40: You shouldn't need froce.

nate-double-u:
OK, testing the `netlify.toml` change:
https://605bb02c277eb10007b716ee--etcd.netlify.app/docs/v2/
https://605bb02c277eb10007b716ee--etcd.netlify.app/docs/v2
(trailing slash, no trailing slash)
Both work.

Next, testing just updating the `index.redirects` file. (leaving this one for test 2, so if it works we can just leave it. Sorry for the PR spam. It looks like `netlify dev` behaves a little differently than Netlify does)

layouts/index.redirects
```diff
@@ -1,3 +1,4 @@
 {{ $latest := site.Params.versions.latest }}
 /docs/latest     /docs/{{ $latest }}
 /docs/latest/*     /docs/{{ $latest }}/:splat
+/docs/v2/*    /docs/v2.3/:splat 301
```
chalin, line 4: You don't need the 301.

nate-double-u:
Testing the `index.redirects` file, both
https://605bb1b1e3f1c200072c1494--etcd.netlify.app/docs/v2
https://605bb1b1e3f1c200072c1494--etcd.netlify.app/docs/v2/

work.

layouts/index.redirects
chalin: And you'll need two rules. One for `/docs/v2` and the other that you have here.

nate-double-u:
we'll squash this PR 😅

layouts/index.redirects
```diff
@@ -1,3 +1,5 @@
 {{ $latest := site.Params.versions.latest }}
 /docs/latest     /docs/{{ $latest }}
 /docs/latest/*     /docs/{{ $latest }}/:splat
+/docs/v2    /docs/v2.3/:splat
```
chalin, line 4: No splat in this rule:

     /docs/v2 /docs/v2.3

layouts/index.redirects
chalin: FYI, AFAIR, rules starting with `/a/b/c` or `/a/b/c/` are the same.

nate-double-u:
OK, i think that's got it

https://deploy-preview-154--etcd.netlify.app/docs/v2
https://deploy-preview-154--etcd.netlify.app/docs/v2/

layouts/index.redirects
```diff
@@ -1,3 +1,5 @@
 {{ $latest := site.Params.versions.latest }}
 /docs/latest     /docs/{{ $latest }}
 /docs/latest/*     /docs/{{ $latest }}/:splat
+/docs/v2    /docs/v2.3
```
chalin, line 4: The rules are working. One last change: add a trailing slash to the end of this rule.

chalin: Oh, and to the rule in line 2 as well :).

chalin (approved):
LGTM 🎉

PR #204, opened 2021-04-08
Trim trailing whitespace in all content files

chalin:
(This will make later diffs cleaner.)

/tag cleanup
/reviewer @nate-double-u

### DIFF
--- a/archetypes/blog.md
+++ b/archetypes/blog.md
@@ -4,6 +4,6 @@ date: {{ dateFormat "2006-01-02" .Date }}
 draft: true
 authors:
 - name: Ada Lovelace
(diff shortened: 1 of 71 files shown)


chalin:
I've done a `git diff` ignoring whitespace, and the generated site files are the same as before.

nate-double-u (approved):

nate-double-u:
> I've done a `git diff` ignoring whitespace, and the generated site files are the same as before.

awesome, in that case i have no problem merging this in. Since that means there isn't anyone using the trailing two whitespaces for formatting in the `.md` files, do we want to edit the `.editorconfige` and enforce whitespace?

PR #227, opened 2021-04-15
Drop About from docsy MVP, and make link checker happy

chalin:
- As agreed via Slack, this removes any link to the new **About** page (while keeping the page for evaluation later when we start bringing in the new IA)
- Eliminate `/latest/` links
  - On the homepage: instead link directly to the param `versions.latest` (this is the `docsy` branch port of #221)
  - In the top-nav: temporarily using `v3.4` instead. I've added a FIXME tag to ensure that we revert this change later.

I can confirm that the link checker is now happy:

```console
$ htmltest
htmltest started at 11:03:28 on public
========================================================================
✔✔✔ passed in 1.417652379s
tested 401 documents
```

/tag docsy
/milestone 21Q2-docsy
/reviewer @nate-double-u

### DIFF
--- a/content/en/_index.html
+++ b/content/en/_index.html
@@ -19,7 +19,7 @@ features:
   A distributed, reliable key-value store for the most critical data of a distributed system
 </h2>
 <div class="mt-5 mx-auto">
-	<a class="btn btn-lg btn-primary mr-3 mb-4" href="/docs/latest/" data-proofer-ignore>
+	<a class="btn btn-lg btn-primary mr-3 mb-4" href="/docs/{{< param versions.latest >}}/" data-proofer-ignore>
 		Docs<i class="fas fa-arrow-alt-circle-right ml-2"></i>
 	</a>
 	<a class="btn btn-lg btn-secondary mr-3 mb-4" href="https://github.com/etcd-io/etcd">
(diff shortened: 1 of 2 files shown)


content/en/_index.html
```diff
@@ -19,7 +19,7 @@ <h2 class="mt-4">
   A distributed, reliable key-value store for the most critical data of a distributed system
 </h2>
 <div class="mt-5 mx-auto">
-	<a class="btn btn-lg btn-primary mr-3 mb-4" href="/docs/latest/" data-proofer-ignore>
+	<a class="btn btn-lg btn-primary mr-3 mb-4" href="/docs/{{< param versions.latest >}}/" data-proofer-ignore>
```
nate-double-u, line 22: do we still need `data-proofer-ignore` if we're using a hugo generated link here? (it was added to make the link checker happy)

chalin: Oh! I was sure that I had removed that. Thanks, I'll take it out.

chalin: Oh, I did it for `master` via #221.

chalin: Thanks for catching it! Fixed

nate-double-u (approved):

Issue #251, opened 2021-04-26
Use YAML for the site config and file front matter


chalin:
Now that we've migrated to docsy, I'd suggest that we convert the config file to YAML. While TOML is great for _really_ simple config files I find it difficult to read when there's lost of nested config like we have. YAML has other features I've found useful on other CNCF sites.

/milestone post-docsy

nate-double-u:
I agree with this -- I find YAML much more human friendly too.

PR #254, opened 2021-04-26
converting config file from toml to yaml

nate-double-u:
Moving file: config.toml -> config.yaml

Updating content from toml format to yaml.

fixes https://github.com/etcd-io/website/issues/251
/cc @chalin


nate-double-u:
I ran a diff on the `public` folder after building `master` and `251-convert-config-to-yaml` branches. No differences found.

chalin (approved):

chalin:
Ahhh, thanks for the conversion and the cleanup! So much more readable (IMHO)!

PR #258, opened 2021-04-27
Updating footer copyright date to match LICENSE

nate-double-u:
Updating footer copyright date to match LICENSE

Contributes to https://github.com/etcd-io/website/issues/247

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -18,7 +18,7 @@
         {{ end }}
       </div>
       <div class="col-12 col-sm-4 text-center py-2 order-sm-2">
-        {{ with .Site.Params.copyright }}<small class="text-white">&copy; 2013-present {{ . }}
+        {{ with .Site.Params.copyright }}<small class="text-white">&copy; 2013-{{ now.Year }} {{ . }}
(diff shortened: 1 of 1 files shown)


layouts/partials/footer.html
```diff
@@ -18,7 +18,7 @@
         {{ end }}
       </div>
       <div class="col-12 col-sm-4 text-center py-2 order-sm-2">
-        {{ with .Site.Params.copyright }}<small class="text-white">&copy; {{ now.Year}} {{ . }}
+        {{ with .Site.Params.copyright }}<small class="text-white">&copy; 2013-present {{ . }}
```
chalin, line 21: I'd vote to keep the dynamically generated year (just because we can in this partials file, whereas we can't in the static LICENSE file), so:

    2013-{{ now.Year }}

chalin (changes requested):
See inlined comment. Thx

layouts/partials/footer.html
nate-double-u: I agree, I'll make that update.

chalin (approved):
👍

PR #259, opened 2021-04-27
Updating author frontmatter format

nate-double-u:
Docsy theme picks up `author` for attribution line. I don't think we'll need to update the template files to sort this issue.

contributes to #247

/cc @chalin


nate-double-u:
deploy preview: https://deploy-preview-259--etcd.netlify.app/blog/2020/jepsen-343-results/

chalin (approved):
I'm all for using the default theme features if we can.

PR #260, opened 2021-04-27
Adding .gitattributes file

nate-double-u:
Adding *.0 - *.9 classification lines so that GitHub categorizes the languages used in the repo more correctly.

I don't expect we'll know if this works until we merge it into master. If it doesn't sort the issue we can revert the merge.

potentially solves #164

/cc @chalin

PR #264, opened 2021-04-27
Rename metrics-list files

### LINKED ISSUE #164: Github incorrectly reports the etcd-io/website repo as mostly composed of Roff files
Screen Shot 2021-03-25 at 1 57 40 PM

I expect our repo is experiencing a similar issue to the one [reported here](https://github.com/github/linguist/issues/3613#issuecomment-302042763) due to all of the files ending in a number in the `/content/docs/vX.Y/metrics` folders.

chalin:
- A proposed solution to / fix #164. An advantage to this approach is that users can view the metrics files directly in their browser (before downloading it if they wish).
- Possible alternative to #260 (which this PR would supersede if accepted)

We could do better than this by displaying the content of the text file in a regular website page (using an appropriate Hugo template), but this seemed like a good enough step forward.

Previews:
- https://deploy-preview-264--etcd.netlify.app/docs/next/metrics/#generated-list-of-metrics
- https://deploy-preview-264--etcd.netlify.app/docs/v3.4/metrics/#generated-list-of-metrics
- ...

/tag cleanup
/reviewer @nate-double-u


nate-double-u:
I think this is a better solution than #260, and I like that it makes the files more usable.

nate-double-u (approved):

PR #274, opened 2021-05-09
dl-build: substitute build.sh for deprecated build

fabjan:
I just pulled latest master and following the directions gave me a deprecation warning:

```shell
$ ./build
\e[91mDEPRECATED!!! Use build.sh script instead.\e[0m
```

I don't know if the deprecation plan involves something special for Windows, or what the old `./build` program did, but the current one just sources `./build.sh` after printing the deprecation warning so that's what's in my proposed change.

### DIFF
--- a/content/en/docs/next/dl-build.md
+++ b/content/en/docs/next/dl-build.md
@@ -14,7 +14,7 @@ The easiest way to get etcd is to use one of the pre-built release binaries whic
 
 ## Build the latest version
 
-For those wanting to try the very latest version, build etcd from the `master` branch. [Go](https://golang.org/) version 1.13+ is required to build the latest version of etcd. To ensure etcd is built against well-tested libraries, etcd vendors its dependencies for official release binaries. However, etcd's vendoring is also optional to avoid potential import conflicts when embedding the etcd server or using the etcd client.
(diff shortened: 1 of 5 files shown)


nate-double-u:
Hi @fabjan, thanks for this!

Would you be able to update the https://etcd.io/docs/next/dl-build/ file as well? It likely suffers from the same issue.

I'll look into why the build failed -- given the changes made i find it a bit surprising that it did.

fabjan:
@nate-double-u Sure thing! I just clicked the "Edit this page" button on the docs and edited directly GitHub (hence the uninformative branch name as well) so did not look much at the context around the change.

Looks like the preview deploy failed because hugo had problems with some vendored bootstrap files, but I don't know anything about the actions setup here.

nate-double-u:
@fabjan, @chalin has fixed the preview issues with https://github.com/etcd-io/website/pull/274. I expect that this PR would have to be rebased in order to trigger a correct build with the new configuration.

As for the edits to the older version pages, I tend to prefer to leave them as historic artifacts and really only change them for factual errors (or converting things like old broken links to their archival versions). For instance removing the unused link is probably OK, but I'd prefer not to change where the body copy links to if the link is still working.

nate-double-u (approved):
Thanks @fabjan!

Issue #149, opened 2021-03-24
Rename master branch to main


chalin:
In conformance to the inclusive naming initiatives of various organizations (including LF/CNCF, GH), `master` should be renamed to `main`. Note that it's been safe to do this for a while now given that GitHub handles the necessary redirects. For details, see:

- [github/renaming][]
- [Links to deleted branches now redirect to the default branch][]

Examples of repos across LF/CNCF that have renamed their default branches include: [grpc.io](https://github.com/grpc/grpc.io) and [grpc-community](https://github.com/grpc/grpc-community).

[github/renaming]: https://github.com/github/renaming
[Links to deleted branches now redirect to the default branch]: https://github.blog/changelog/2020-07-17-links-to-deleted-branches-now-redirect-to-the-default-branch

cc @celestehorgan @nate-double-u

nate-double-u:
@spzala, @chalin and I were thinking that since we've gotten the Docsy conversion in, that we might do this branch rename soon. I'm not able to make the branch name change, so I've assigned the issue to you.

chalin:
@spzala - a friendly ping, since it's been two weeks: might you be able to perform the rename or give @nate-double-u or me the necessary permissions?

gyuho:
Just made @chalin and @nate-double-u admin for this repository. Please feel free to move forward with the renaming.

spzala:
@gyuho thanks! @chalin sorry for the delay, got caught up with Kubecon activities last week.

chalin:
Done!

FYI, here is what GitHub had to say:

> Renaming this branch: will update 8 pull requests targeting this branch across 8 repositories.

**If you have a _fork_** of this repo, you should visit your fork and rename the default branch to `main` as well -- in fact GitHub will invite you to do so.

**If you have a _local clone_**, you can update it by running the following commands provided by GitHub (this will work for most folks):

```console
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```

chalin:
Netlify is now also deploying from `main`:

> <img width="724" src="https://user-images.githubusercontent.com/4140793/117999965-64185980-b313-11eb-9182-32eb88a698f9.png">

PR #281, opened 2021-05-12
Delete unused nav-panel.html

chalin:
- Contributes to #247
- The file is a relic from the pre-docsy days.
- There are no changes in the generated site files.

This is part of a more general post-docsy cleanup that needs to be done. I'm only doing this single file now because it is in support of #282 (and hence #149).


chalin:
This is part of a more general post-docsy cleanup that needs to be done. I'm only doing this single file now because it is in support of #282 (and hence #149).

nate-double-u (approved):

PR #286, opened 2021-05-17
Update libraries and tools

suizman:
Include mingchuno/etcd4s library for Scala with V3 support and watch using Akka Stream.

### DIFF
--- a/content/en/docs/v3.4/integrations.md
+++ b/content/en/docs/v3.4/integrations.md
@@ -41,7 +41,7 @@ description: A listing of etcd tools and client libraries
 
 - [maciej/etcd-client](https://github.com/maciej/etcd-client) - Supports v2. Akka HTTP-based fully async client
 - [eiipii/etcdhttpclient](https://bitbucket.org/eiipii/etcdhttpclient) - Supports v2. Async HTTP client based on Netty and Scala Futures.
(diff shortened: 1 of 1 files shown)


nate-double-u (changes requested):
Hi @suizman, thanks for this.

We're keeping the older versions mostly as an archive, and try not to change them with new information or links (beyond fixing broken things). Is this info relevant to the `/next` (v3.5) or `/v3.4` versions? If so, could you move this info there?

nate-double-u:
I've opened [an issue](https://github.com/etcd-io/website/issues/287) regarding the build failure. I'm not yet sure what's causing it, but it looks similar to another recent, failing PR.

nate-double-u:
@suizman, @chalin has fixed the preview issues with #274. I expect that this PR would have to be rebased in order to trigger a correct build with the new configuration.

suizman:
Hi @nate-double-u, my bad changing the docs for an older version. Now the change is at the correct version **/v3.4**

chalin:
@suizman - I tweaked the description, hopefully that's ok with you.

chalin (approved):

chalin:
This change needs to be propagated to `/docs/next`, would you be willing to make that change @suizman? I'll open an issue to track this.

PR #292, opened 2021-05-20
Package.json cleanup: remove unused bulma et al

chalin:
- Contributes to #247
- There is no change in the generated site files


nate-double-u (approved):
LGTM

PR #288, opened 2021-05-18
Installation page for etcd / etcdctl (update from dl-build)

nate-double-u:
https://github.com/etcd-io/website/issues/267, https://github.com/etcd-io/website/pull/278#discussion_r632813473

Note, merge https://github.com/etcd-io/website/pull/278 first.

First pass at refactoring dl-build into an installation page.

Currently, the only relevant file is `content/en/docs/next/install.md`

Deploy preview: https://deploy-preview-288--etcd.netlify.app/docs/next/install/

PR #293, opened 2021-05-20
Download pages: line-wrap markdown

chalin:
- Only line-wrapped some paragraphs. This will make it easier to review changes that are coming (like #288).
- No change in content.


nate-double-u (approved):
LGTM

PR #294, opened 2021-05-20
Installation page (docs/next) from reworked download page

chalin:
- These updates streamline the Download page -- keeping it [DRY][] relative to the full site docs.
- Contributes to #318

Preview: https://deploy-preview-294--etcd.netlify.app/docs/next/dl-build/

@spzala @ptabor @gyuho PTAL -- btw, are there plans to support installation using `go install`?

[DRY]: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

### DIFF
--- a/content/en/docs/next/dl-build.md
+++ b/content/en/docs/next/dl-build.md
@@ -1,83 +1,77 @@
 ---
-title: Download and build
+title: Install etcd
+linkTitle: Install
 weight: 1150
-description: Instructions for downloading and building different versions of etcd
+description: Instructions for installing etcd from pre-built binaries or from source.
 ---
 
-## System requirements
+## Requirements
 
-The etcd performance benchmarks run etcd on 8 vCPU, 16GB RAM, 50GB SSD GCE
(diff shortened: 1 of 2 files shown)


nate-double-u:
Deploy preview: https://deploy-preview-294--etcd.netlify.app/docs/next/dl-build/

nate-double-u (approved):
This looks good to me. I like that you're omiting `etcdctl` and pointing back to the Quickstart.

I'd be happy to close https://github.com/etcd-io/website/pull/288 in favour of this PR.

Next steps questions
* should this file be renamed and a forward added?
* should this file be duplicated into `v3.4`?

chalin:
@nate-double-u: yes the plan is to rename the file and add a redirect rule, but only in a followup PR. Ditto for v3.4 -- let's get this version approved first so that we don't have to make edits in two places.

chalin:
Hi all: we're coming on a week tomorrow for this PR, so if there are no objections by EOD today, I'll assume that this is good to go. If you need more time, let me know.

Issue #266, opened 2021-04-30
can we lowercase "etcd" uniformly in the document?


tangcong:
can we lowercase etcd uniformly in the document, so that it looks more beautiful and standardized? thanks.
![image](https://user-images.githubusercontent.com/2403673/116643477-bff1f280-a9a3-11eb-9a93-9e98d625766f.png)

nate-double-u:
There are some instances of "Etcd" in the text which should probably be made lower case, but for the header I'd expect this to be a scss change somewhere.

PR #299, opened 2021-05-21
Setting etcd to lowercase throughout the docs

nate-double-u:
Updating copy to use lowercase "etcd" where needed.

contributes to: #266 (logo in navbar still needs to be addressed)


layouts/partials/navbar.html
```diff
@@ -0,0 +1,33 @@
+{{ $cover := .HasShortcode "blocks/cover" }}
+<nav class="js-navbar-scroll navbar navbar-expand navbar-dark {{ if $cover}} td-navbar-cover {{ end }}flex-column flex-md-row td-navbar">
+        <a class="navbar-brand" href="{{ .Site.Home.RelPermalink }}">
+		<span class="navbar-logo">{{ if .Site.Params.ui.navbar_logo }}{{ with resources.Get "icons/logo.svg" }}{{ ( . | minify).Content | safeHTML }}{{ end }}{{ end }}</span><span class="text-lowercase font-weight-medium">{{ .Site.Title }}</span>
```
nate-double-u, line 4: Only difference between this and the original Docsy file is:

`<span class="text-lowercase font-weight-medium">`

nate-double-u:
Deploy preview:  https://deploy-preview-299--etcd.netlify.app/

chalin (changes requested):
LGTM, except that I'd handle the navbar fix separately. Can you pull out that change? I'd rather avoid copying an entire partial for that if it can be avoided.

chalin (approved):

PR #303, opened 2021-05-21
Lowercase etcd in the navbar

### LINKED ISSUE #266: can we lowercase "etcd" uniformly in the document?

chalin:
Closes #266

### DIFF
--- a/assets/scss/_styles_project.scss
+++ b/assets/scss/_styles_project.scss
@@ -129,3 +129,10 @@
 .bg-gray-100 {
   background-color: $gray-100;
 }
+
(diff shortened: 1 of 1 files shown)


nate-double-u (changes requested):
Are we able to use this method to bring down the bolding a bit? 'etcd' is quite bold, and while it isn't normally an issue, with the actual logo right below it on the home page it looks a bit off. I think a medium weight may be better.

chalin:
I'd prefer keeping the bolding level as is for now (at least in the context of this PR).

nate-double-u (approved):

Issue #267, opened 2021-05-04
New IA implementation


chalin:
Reference: #110

chalin:
@nate-double-u: good first pass. Here are some suggestions for the "sitemap" (which isn't really a sitemap -- maybe we should call it "Permalink structure"):

- [ ] Drop the "homepage" node, and left-shift all entries below it. We know everything is rooted below the homepage, and besides, we have a separate section for discussing the "homepage"
- [ ] Add the permalink path name to each node: e.g. Docs (`/docs`) -- yes, it's trivial for "Docs" but less so for entries below that.
- [ ] Consider the shorter name "Get started" rather than "Getting started"
- [ ] Why would "Administration", "Get started", etc not be versioned?

nate-double-u:
> Why would "Administration", "Get started", etc not be versioned?

Good question. I think I will move them under the versioning. 

Also, what about pages like /docs/v3.4/triage(/issues), or /docs/v3.4/upgrades? Should they be removed from under versioning? I expect that the process of opening a ticket won't change from version to version. Upgrades is a list of X.Y to X.Z upgrades, so is mostly duplicated content.

nate-double-u:
While we go through this process we should consider standardizing "-" vs "_" in links.

chalin:
So in a nutshell, based on the reading of the docs I’ve done this week (05/09), here’s what I’m feeling needs to be done:

- Finalize a **Get started** page (considering the ectd README version, possibly).
  Then link to that page from the homepage (which will mean that we need a v3.4 version)
- Consider (re-)writing an **Installation** page, maybe using the download/build page as a base (but remove the testing part -- we can direct folks to the get-started page).
  We might also want to add that to the top-nav
- Consider reworking the demos into **tutorials**
- Consider writing a **Community** page (and adding a link to it in the top-nav). The info can come from the etcd README, in particular.
- Consider adding a link to [**play.etcd.io**](http://play.etcd.io/) to the homepage and/or top-nav

All of this is open to discussion, of course.

nate-double-u:
I had initially tried to [organize the IA](https://github.com/etcd-io/website/issues/110) into an _Evaluation, Addition, Fix, and Features_ model; would a _Concepts, Tasks, Tutorials, and Reference_ model (like on the [k8s.io/docs](https://kubernetes.io/docs/home/) page) be a better model to go with given what we've got already built out?

nate-double-u:
[...]

PR #305, opened 2021-05-22
Homepage CTA buttons: replace GH by Learn more

chalin:
- CTA buttons should be just that, call-to-action. We have a link to GH in the footer of every page, so we don't need it to take up a CTA button slot. Instead it makes more sense to link to the latest docs. This PR makes that change and switches the order of the buttons so that Quickstart is second (but keeps its icon so it is more eye-catching the "Learn more" before it).
- In the second section of the page, now that we have a "Learn more" in the hero section, it makes sense to have the "Learn more" that appears there be less prominent (saving homepage vertical real estate).
- UX / navigation contribution to #267

### DIFF
--- a/content/en/_index.html
+++ b/content/en/_index.html
@@ -19,11 +19,11 @@ features:
   A distributed, reliable key-value store for the most critical data of a distributed system
 </h2>
 <div class="mt-5 mx-auto">
+	<a class="btn btn-lg btn-secondary mr-3 mb-4" href="/docs/{{< param versions.latest >}}/">
+		Learn more
+	</a>
 	<a class="btn btn-lg btn-primary mr-3 mb-4" href="/docs/{{< param versions.latest >}}/quickstart">
 		Quickstart<i class="fas fa-arrow-alt-circle-right ml-2"></i>
-	</a>
-	<a class="btn btn-lg btn-secondary mr-3 mb-4" href="https://github.com/etcd-io/etcd">
-		GitHub <i class="fab fa-github ml-2 "></i>
 	</a>
 </div>
 {{< /blocks/cover >}}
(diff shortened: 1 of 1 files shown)


nate-double-u (approved):
I agree with the change to the CTA buttons. One small nit, do we want to find another, appropriate, icon for the _Learn More_ button?

It's a small change, but I appreciate the space saving on the home page. It brings the page together without crowding it.

chalin:
Most other sites I know don't have such icons. I only included it here to catch-the-eye for the Quickstart button, which is now second.

If you feel strongly about consistency, I'd prefer dropping the icon for the Quickstart button.

nate-double-u:
Given the choice, I think I'd rather have the eye-catch than consistency in this case.

hexfusion:
I dont have a problem with this directly but GH link should be very easy to access. It feels a little hidden now on the bottom.

chalin:
> I dont have a problem with this directly but GH link should be very easy to access. It feels a little hidden now on the bottom.

True (that the links to GH aren't prominent), and that is by design. From a documentation-website design / UX point of view, it is quite conventional to have links to the GH repo in the footer, rather than a Call-To-Action (CTA) button. CTA buttons are precious real estate used to direct readers to where they (most of them) need to go. For example, see https://kubernetes.io/. 

Looking at Google Search Console results, the most popular etcd search is "etcd", which returns etcd.io followed by the link to the etcd repo, as top hits. The assumption is that if a reader clicks on the docs site, they want docs, not the repo. If they want the repo, then it's the second result in Google Search (for example) and the assumption is that they'd have clicked that link directly. Of course, there are other ways that readers can reach the site, but let's stop there for now.

Unless you feel strongly about this, I'd suggest we keep it as is. Still, thanks for the feedback @hexfusion.

chalin:
@hexfusion - I'm not against making the link to GH more prominent, though it might be easier to assess the best way to do so once we've landed more of the new IA (#267). If you still feel that this is worth fighting for :), ... could you please open an issue? Thanks

PR #324, opened 2021-05-26
Netlify previews: also build draft and future pages

chalin:


chalin:
This is (hopefully) temporary, until v3.5 lands. #325 is the followup issue. /cc @nate-double-u

PR #327, opened 2021-05-26
Disable CI link checking until we get it fully working

chalin:
(If we don't do this, v3.5 links will be considered broken because the current link checker doesn't process redirects.)


nate-double-u (approved):
LGTM

PR #338, opened 2021-05-28
Always show Versions menu, though disabled when necessary

### LINKED ISSUE #304: Versions menu UX issue
My original thought was that the Versions menu shouldn't take up real estate in the top nav if it isn't being shown. For example:

Screen Shot 2021-05-21 at 2 27 31 PM

Screen Shot 2021-05-21 at 2 27 50 PM

nate-double-u:
Now we always display the **Versions** menu but have made it **inactive** on pages where it doesn't apply.

fixes #304.

### DIFF
--- a/layouts/partials/navbar-version-selector.html
+++ b/layouts/partials/navbar-version-selector.html
@@ -25,4 +25,8 @@
       <a class="dropdown-item" href="{{ $targetURL }}">{{ $versName }}</a>
     {{ end -}}
   </div>
+{{ else }}
+  <a class="nav-link dropdown-toggle disabled" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+    {{- .Site.Params.version_menu -}}
+  </a>
 {{ end -}}


nate-double-u:
Deploy preview: https://deploy-preview-338--etcd.netlify.app/

chalin (changes requested):
Thanks. Let's refactor this so that the dropdown html is always present, but the anchor has the added `disabled` class when not under `/docs`.

nate-double-u:
> Thanks. Let's refactor this so that the dropdown html is always present, but the anchor has the added `disabled` class when not under `/docs`.

I'm not sure how we can (if I'm understanding this request correctly).

If you mean something like
`<a class="nav-link dropdown-toggle {{ if ne .Section "docs" -}}disabled{{ end -}}" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">`
instead of the if else structure I've put in around the anchor/dropdown, then the template fails because the dropdown menu is looking for front-matter in the docs files that isn't present in the non-versioned pages.

I'm not sure what the dropdown html should be building on non versioned pages.

chalin:
> If you mean something like

Yes, almost: move the space inside the if: `<a class="nav-link dropdown-toggle{{ if ne .Section "docs" -}} disabled{{ end -}}"`

> the template fails

Darn. I wasn't expecting useful dropdown entires, but also wasn't expecting it to fail.

Ok then, put the `if eq .Section "docs"` around the _content_ of the `<div class="dropdown-menu" ...></div>`. Thanks.

nate-double-u:
> Ok then, put the `if eq .Section "docs"` around the _content_ of the `<div class="dropdown-menu" ...></div>`. Thanks.

The last commit does this -- but I don't like it as much as having a visually distinct _disabled_ versions selector.

nate-double-u:
> Ok then, put the `if eq .Section "docs"` around the _content_ of the `<div class="dropdown-menu" ...></div>`. Thanks.

The last (force pushed) commit does this -- but I don't like having an empty dropdown as much as having a visually distinct disabled versions selector.

chalin (changes requested):
> I don't like having an empty dropdown as much as having a visually distinct _disabled_ versions selector.

Right, neither do I. We need both. So you can add the conditional inclusion of `disabled` (the way you had suggested earlier).

nate-double-u:
> ... We need both. So you can add the conditional inclusion of `disabled` (the way you had suggested earlier).

OK, added -- on Firefox disabling doesn't allow the dropdown to work, so we don't get an empty one on click. waiting for the build to check other browsers.

nate-double-u:
wait, i didn't do that quite correctly.
give me a moment :)

nate-double-u:
OK, the latest commit gives the `dropdown-menu` a chance to be empty if it's able to be shown at all.

chalin:
@nate-double-u - On Friday I was on the fence between accepting your original change and asking for extra rework in the context of this PR or later. I've changed my mind and, for now at least, I prefer the simplicity of your original change -- so I went ahead and stripped the extra commits to leave only the first one. While there is an opportunity to rework this fix to keep things a bit more DRY, that can come later.

chalin (approved):

Issue #296, opened 2021-05-21
Versions menu: visually distinguish page vs. version-home targets


chalin:
If you're on page P, the versions menu will contain links to P in all other versions that have such a page. When there is no such page for version vX.Y then ` /docs/vX.Y` is used as a target. Currently, there is no visual cue to distinguish between these, but there should be.

nate-double-u:
I find myself hovering over the menu to see if there is another version of the current page. A visual distinction would be helpful.

kapunahelewong:
I can help out here!

nate-double-u:
Some pics to clarify:

On page https://etcd.io/docs/v3.5/quickstart/

![Screen Shot 2021-06-22 at 10 11 24 AM](https://user-images.githubusercontent.com/4453979/122970092-fa14ba80-d37c-11eb-9ebb-3ec6dad67737.png)
Note the subtle highlighting when hovering over a version. Also note the link in the lower left hand corner -- there is a v3.4 version of this page, so the menu links to that page.

![Screen Shot 2021-06-22 at 10 11 28 AM](https://user-images.githubusercontent.com/4453979/122970087-f97c2400-d37c-11eb-8018-b419108a8566.png)
Note the link in the lower left again, v3.3 doesn't have a version of the Quickstart page, so the menu links to the root of v3.3 instead.

The code that builds this drop-down menu is here: https://github.com/etcd-io/website/blob/main/layouts/partials/navbar-version-selector.html

PR #390, opened 2021-06-22
add styles to indicate existence of doc for each version

kapunahelewong:
Contributes to #296

Adds icons showing that a doc exists or that a doc doesn't for each version. Used an icon for accessibility in case of color blindness or other visual issues where color isn't a sufficient indicator.

## Hover state and icon for non-existent doc
Screen Shot 2021-06-22 at 3 53 28 PM

## Hover state and icon for existent doc
Screen Shot 2021-06-22 at 3 53 32 PM

Another option is that we dispense with the no-doc icon and keep the file icon. That might be enough. Also, maybe there's a better way to select the href attribute allowing for variance.

### DIFF
--- a/assets/scss/_styles_project.scss
+++ b/assets/scss/_styles_project.scss
@@ -216,3 +216,56 @@
   padding: 0 3px;
   font-size: .7rem;
 }
+
+// ************************************
+
+// Styles for icon in version menu
+
+// ************************************
+
+// style options in Versions drop down on the upper right of the docs so that it's clear if a doc exists for a particular version as not all docs exist in all versions
+
+//  establish fontawesome on dropdown:after
+
+a.dropdown-item:after {
+
+  font-family: 'Font Awesome\ 5 Free';  
+  position: relative;
+  bottom: 1px;
+  padding: 0 3px;
+  font-size: .7rem;
+
+}
+
+// Document icon is from https://fontawesome.com/v5.15/icons/file-alt?style=solid and signifies a doc exists in a particular version
+
+a.dropdown-item[href*="/docs/"] {
+  
+  &:hover {
+    background: #a0d0ff linear-gradient(180deg, #1f6897, #1f6897) repeat-x;
+    color: #fff;
+  }
+  
+  &:after {
+    content: "\f15c";
+  }
+}
+
+// Remove the doc icon from the version numbers in the dropdown that don't have a corresponding doc. 
+// ~= means that the href value ends in the string specified. Maintainers will have to add new versions as they come out if there aren't docs for that version.
+a.dropdown-item[href~="/docs/v2.3/"], 
+a.dropdown-item[href~="/docs/v3.1/"], 
+a.dropdown-item[href~="/docs/v3.2/"], 
+a.dropdown-item[href~="/docs/v3.3/"], 
+a.dropdown-item[href~="/docs/v3.4/"],
+a.dropdown-item[href~="/docs/v3.5/"] {
+
+  &:hover {
+    background:#accff3 linear-gradient(180deg, #deeaf7, #e3ecf5) repeat-x;
+    color: #222;
+  }
+  
+  &:after {
+    content: "";
+  }
+}


kapunahelewong:
Preview: https://deploy-preview-390--etcd.netlify.app/docs/v3.5/quickstart/

nate-double-u (commented):
Thanks for this @kapunahelewong!

> Used an icon for accessibility in case of color blindness or other visual issues where color isn't a sufficient indicator.

That's a good thought, and I think that only using a highlight may not communicate the distinction well enough alone.

> Another option is that we dispense with the no-doc icon and keep the file icon.

This may be a good solution. I'll take a look around the free font awesome and see if there is a better icon as well.

Could you bring up the visibility of the nope highlighting? It's still very subtle and may be missed.

celestehorgan:
@nate-double-u tagged me to look at this from a UX perspective. 

Rather than (or in addition to) an inactive link, could we apply a class to versions where the file doesn't exist that displays the link in a lighter gray?

nate-double-u:
> Rather than (or in addition to) an inactive link

It's not an inactive link per se. When there is no matching file in another version, the menu links to that versions root rather than being inactive.

nate-double-u:
[This icon](https://fontawesome.com/v5.15/icons/file-excel?style=regular) may be a good alternative to the circle and bar (if we decide to go with a no-doc option). It's an excel logo, but taken out of context, it may work well at indicating there isn't a match in the other version.

spzala (commented):
@kapunahelewong @nate-double-u  lgtm overall, however I see  case where all circle icon  (non-exist) for all the versions, so may want to fix it (link below). Also, I am good with the icons that @nate-double-u suggested but I don't have presences, so I would go with what you and @nate-double-u decide.  Thanks! 

image

kapunahelewong:
What do you think of this blue for the nope hover state? (I checked the contrast ratio and it passes too, btw). I also removed the nope icon to see how it looks. It's cleaner and I think still conveys the message. Here's a screenshot:

Screen Shot 2021-06-23 at 9 34 20 AM


Also, out of curiosity, I tried putting words there but it starts getting cluttered. 
Screen Shot 2021-06-23 at 9 44 20 AM

 So I tried them on hover. This might be getting interesting (with better copy/iconography):
Screen Shot 2021-06-23 at 9 49 31 AM

The preview is of the first screenshot: https://deploy-preview-390--etcd.netlify.app/docs/v3.5/quickstart/

kapunahelewong:
> @kapunahelewong @nate-double-u lgtm overall, however I see case where all circle icon (non-exist) for all the versions, so may want to fix it (link below). Also, I am good with the icons that @nate-double-u suggested but I don't have presences, so I would go with what you and @nate-double-u decide. Thanks!
> 
> image

Thank you so much, @spzala for catching that. I think I've fixed it with this push, but I wonder then if we can't do the hover nope states like I was thinking in https://github.com/etcd-io/website/pull/390#issuecomment-866854388. Hm, I need a better selector if we want to go that way.

kapunahelewong:
> [This icon](https://fontawesome.com/v5.15/icons/file-excel?style=regular) may be a good alternative to the circle and bar (if we decide to go with a no-doc option). It's an excel logo, but taken out of context, it may work well at indicating there isn't a match in the other version.

Oops - it seems my page hadn't refreshed with everyone's comments. Pardon my oddly timed responses, everyone!
@nate-double-u I really like that excel icon for the nope state! Let me see if I can refine that selector so they don't apply at the higher levels like @spzala noticed.

nate-double-u:
> What do you think of this blue for the nope hover state? (I checked the contrast ratio and it passes too, btw). 

I like this blue combo for the yup/nope states.

> could we apply a class to versions where the file doesn't exist that displays the link in a lighter gray?

I think this may be good for when the menu isn't in a hover state.

> I also removed the nope icon to see how it looks. It's cleaner and I think still conveys the message. 

I almost want to try the reverse of this -- only have an icon for the nope state. I think this may make the menu less cluttered in more places (with the exception of v2.3, most pages should have analogs in other versions). I think that'll make the docs root that @spzala noticed a bit more complex though.

kapunahelewong:
> I almost want to try the reverse of this -- only have an icon for the nope state. I think this may make the menu less cluttered in more places (with the exception of v2.3, most pages should have analogs in other versions). I think that'll make the docs root that @spzala noticed a bit more complex though.

Yeah, I can switch it so the nope state is the only one with the icon, but then the selector isn't specific enough to differentiate between this page https://etcd.io/docs/v3.5/ and, for example, this: https://etcd.io/docs/v3.5/quickstart/
so we get those errant nope icons on https://etcd.io/docs/v3.5/.

Is there a way to add a class to a parent that's in the DOM only when on a page that the nope state should be applied to the dropdown?

chalin (approved):
Thanks @kapunahelewong!

@nate-double-u - I'd vote to merge this as is. I can then make followup changes (which will require updates to partials), after. WDYT?

nate-double-u:
I agree @chalin!
/approve

nate-double-u (approved):

chalin:
@nate-double-u - this functionality seems to have been lost / removed. Any idea why?

PR #392, opened 2021-06-23
Adding Google Search

nate-double-u:
Adding Google Search by the [Docsy theme method](https://www.docsy.dev/docs/adding-content/navigation/#configure-search-with-a-google-custom-search-engine).

Deploy preview: https://deploy-preview-392--etcd.netlify.app/

fixes: https://github.com/etcd-io/website/issues/340

_Note_: Since there is an alternate solution being proposed, this PR should be closed if https://github.com/etcd-io/website/pull/403 is chosen as the better solution.

PR #403, opened 2021-07-02
Adding Lunr Search

nate-double-u:
Adding Lunr.js search via the [Docsy theme](https://www.docsy.dev/docs/adding-content/navigation/#configure-local-search-with-lunr).

Note, with this search, I was able to add `exclude_search: true` to any deprecated versions so that only the most recent version shows up in the search. Each version can have this set individually, so we can add or remove any version (or page in the `content/` folder) we want from search results.

Deploy preview: https://deploy-preview-403--etcd.netlify.app/

Alternate fix for: https://github.com/etcd-io/website/issues/340

_Note_: Since this is an alternate solution, this PR should be closed if https://github.com/etcd-io/website/pull/392 is chosen as the better alternative.

### DIFF
--- a/assets/js/offline-search.js
+++ b/assets/js/offline-search.js
@@ -1,0 +1,198 @@
+// etcd-docsy full file override: we're not tracking changes to the docsy file of the same name.
+
+// Adapted from code by Matt Walters https://www.mattwalters.net/posts/hugo-and-lunr/
+
+(function ($) {
+    'use strict';
+
+    $(document).ready(function () {
+        const $searchInput = $('.td-search-input');
+
+        //
+        // Options for popover
+        //
+
+        $searchInput.data('html', true);
+        $searchInput.data('placement', 'bottom');
+        $searchInput.data(
+            'template',
+            '<div class="popover offline-search-result" role="tooltip"><div class="arrow"></div><h3 class="popover-header"></h3><div class="popover-body"></div></div>'
+        );
+
+        //
+        // Register handler
+        //
+
+        $searchInput.on('change', (event) => {
+            render($(event.target));
+
+            // Hide keyboard on mobile browser
+            $searchInput.blur();
+        });
+
+        // Prevent reloading page by enter key on sidebar search.
+        $searchInput.closest('form').on('submit', () => {
+            return false;
+        });
+
+        //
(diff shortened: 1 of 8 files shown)


caniszczyk:
LGTM as a good start

caniszczyk:
LGTM as a good start, may need some tweaking

nate-double-u:
Thanks -- right off the bat I'd like to fix the `Search results` css -- the `x` is too close to the copy.

nate-double-u:
The last force-push overrides Docsy's `assets/js/offline-search.js` file to add better white space to the results pane.

assets/js/offline-search.js
```diff
@@ -0,0 +1,196 @@
...
+            $html.append(
+                $('<div>')
+                    .css({
+                        display: 'flex',
+                        justifyContent: 'space-between',
+                        marginBottom: '1em',
+                    })
+                    .append(
+                        $('<span>')
+                            .text('Search results\xa0') // \xa0 is the Non-breakable space char (&nbsp;)
```
nate-double-u, line 129: This is the only line that is different from the Docsy `assets/js/offline-search.js` file. I'm not sure I'm happy with overriding a whole file for such a small change.

@chalin, @kapunahelewong, do you have any ideas for a better method of making this fix?

kapunahelewong: oof. Yeah, agreed. I don't know though. I'll keep an eye out for a solution. Does the build tree shake duplicate code? If so then it would at least mitigate duplication.

spzala (approved):
lgtm
Thanks @nate-double-u

assets/js/offline-search.js
chalin: For now, add the usual opening line comment stating that it is a Docsy file override.

chalin (approved):
LGTM (though I haven't tested it)

assets/js/offline-search.js
nate-double-u: Added the line.

chalin: > @chalin ... do you have any ideas for a better method of making this fix?

After looking at this more closely, yes 😄 -- see #493.

PR #404, opened 2021-07-15
Update npm packages to avoid high severity vulnerabilities

chalin:
There is no change to the generated site files.

cc @nate-double-u


nate-double-u (commented):
Looks good to me, thanks @chalin!

PR #408, opened 2021-07-16
Makefile: drop obsolete target

chalin:
cc @nate-double-u


chalin:
Build is green, merging.

PR #409, opened 2021-07-16
Netlify config: drop unnecessary x-robots-tag

chalin:
We no longer need the X-Robots-Tag in the branch deploy headers, since Docsy (or Hugo?) manages that based on the build kind (production or development).

Preview: https://deploy-preview-409--etcd.netlify.app/

If you look at the page source, each page of the preview contains the following:

```html
<meta name=ROBOTS content="NOINDEX, NOFOLLOW">
```

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -10,14 +10,3 @@ command = "npm run build:preview"
 
 [context.branch-deploy]
 command = "npm run build:preview"
-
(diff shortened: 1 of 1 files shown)


nate-double-u (approved):
Agreed, this is no longer needed.

Issue #139, opened 2021-03-23
Check for obsolete files after initial docsy-conversion


nate-double-u:
Question about whether or not we need some files still (some came in on a recent rebase)

Can we remove the following (check for yes, we can remove them):
- `layouts/_default/_markup/render-link.html`. Keep, but fix: #270
- [x] `layouts/partials/css.html` #388
- [ ] `layouts/robots.txt`
- `layouts/404.html`, keep as is ~(should we use the docsy default?)~
- [x] `package-lock.json`
- [x] `static/apispec/swagger` (the whole folder, @spzala?) #388

Positive answers to this (i.e. the file should be removed) doesn't have to stop this PR from being merged, we can just open an issue against the `docsy` branch if we need to take action.

_Originally posted by @nate-double-u in https://github.com/etcd-io/website/issues/112#issuecomment-805297501_

nate-double-u:
Once https://github.com/etcd-io/website/pull/388 is merged in we can probably close this issue.

`layouts/_default/_markup/render-link.html` is being managed with https://github.com/etcd-io/website/issues/270, and the other files are either removed with #388, or are likely still needed.

chalin:
`layouts/robots.txt` can be removed. After that, this issue can be closed.

PR #416, opened 2021-07-26
Cleanup of partials that are no longer used under docsy theming

chalin:
- Contributes to #139 
- No change in the resulting generated site files


nate-double-u (approved):
/lgtm
(Confirmed, no changes to built site)

PR #388, opened 2021-06-21
Pruning unneeded static/apispec/swagger and css partial

nate-double-u:
Files removed:
* duplicates
  * static/apispec/swagger/rpc.swagger.json
  * static/apispec/swagger/v3election.swagger.json
  * static/apispec/swagger/v3lock.swagger.json
* not used in build (no diff in public once built)
  * layouts/partials/css.html

Contributes to #139

PR #417, opened 2021-07-26
Use head-css.html docsy override with tweak for IsProduction

chalin:
- This introduces a proper docsy override of the partial formerly named `css.html` (that was deleted via #388).
- Generates a production version based on `hugo.IsProduction` rather than `.Site.IsServer`. This makes it easier to create local development builds.
- No change in the generated site files.

Signed-off-by: Patrice Chalin <pchalin@gmail.com>


nate-double-u (approved):
/lgtm
confirmed no change to production built files.

Issue #419, opened 2021-07-28
Fix links that rely on Hugo's Page.Get function


chalin:
Fixing #270 first requires that we cleanup links that depend on the path rewrites being done by the current render-link hook. This issue is to track the updates to the doc pages, split by version:

- [x] v2.3 pages - #451
- [x] v3.1 pages - #450
- [x] v3.2 pages - #454
- [x] v3.3 pages - #448, and some fixes via #450
- [x] v3.4 pages - #458
- [x] v3.5 pages - #457
- [x] Other pages - #421 

A few more guidelines when fixing broken links:

- Favor the use of relative links when the page being linked to is under the same language version section
- End all paths with a `/`. For refs with anchors, the `/` needs to be placed at the end of the path before the anchor.

chalin:
Hey @nate-double-u and @kapunahelewong: if ever you're in a Fix-it-Friday mood, maybe you'd be willing to correct the links for (some of) the pages of one of the language versions? If so, that would be grand! No obligation, of course -- I know you have other tasks on your plate.

If you'd like to give it a try, then locally apply #420, and then run the link checker: it'll let you know which links are broken. When you submit your PR, only submit the link fixes, not #420.

To ensure that we don't duplicate work, add your GH id next to the language version in the original comment, or just add a new comment stating which language version you'll be working on. Thanks!

kapunahelewong:
Hey @chalin, thank you for the instructions on getting started. How do we run the link checker? 
And would our PRs be like #421?

chalin:
> How do we run the link checker?

See the instructions (and the Note) in [Build or serve the site](https://github.com/etcd-io/website#build-or-serve-the-site).

> And would our PRs be like #421?

Yes. Thanks!

kapunahelewong:
I can start with v3.3.

chalin:
@kapunahelewong is working on v3.1

nate-double-u:
I could take v3.2

nate-double-u:
As per our conversation, @kapunahelewong, I leave it to you 🙂

kapunahelewong:
I'm working on v2.3 now.

kapunahelewong:
Now working on v3.2.

kapunahelewong:
Now I'm working on v.3.5.

kapunahelewong:
Currently working on 3.4! ⭐️

kapunahelewong:
[...]

PR #421, opened 2021-07-28
Misc link updates

chalin:
- Contributes to #419
- No "net" change in the generated site files, and the link checker continues to be happy with the updated link


nate-double-u (approved):
/lgtm

PR #424, opened 2021-07-29
Use the hugo-extended package

chalin:
Through this PR we're adopting the use of the [hugo-extended](https://www.npmjs.com/package/hugo-extended) package, a "Plug-and-play binary wrapper for Hugo Extended".

By doing so we **reduce the burden on contributors**: the _right_ version and variant (_extended_ edition), gets fetched automatically when `npm install` is run. Also, from now on, every version of the website will be associated to the version of Hugo needed to build it, and npm will fetch the proper version on an install.

Now the only prerequisite to building the site is Node.js. I've made corresponding updates to the README.

/cc @kapunahelewong

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,6 @@ manage Node versions:
 
 ```console
 $ nvm install --lts
-$ nvm use --lts
 ```
 
 ## Setup
--- a/README.md
+++ b/README.md
@@ -1,60 +1,62 @@
 # etcd.io
 
-This repository houses all of the assets used to build the etcd docs and website available at https://etcd.io.
+The [etcd.io][] website, built using [Hugo][] and hosted on [Netlify][].
 
-## Run the site locally
+## Build prerequisites
 
-### Prerequisites
-
-To build and serve the site, you'll need these tools:
-
-- **[Hugo, extended edition][hugo-install]**; match the version specified in
-  [netlify.toml](netlify.toml)
-- **Node**, the latest [LTS release][]. Like Netlify, we use **[nvm][]**, the
-  Node Version Manager, to install and manage Node versions:
-  ```console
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,61 +1,63 @@
...
 
-### Prerequisites
+To build and serve the site, you'll need the latest [LTS release][] of **Node**.
+Like Netlify, we use **[nvm][]**, the Node Version Manager, to install and
+manage Node versions:
 
-To build and serve the site, you'll need these tools:
+```console
+$ nvm install --lts
+$ nvm use --lts
```
nate-double-u, line 13: Does `nvm install --lts` always result in the latest node being used? (I've tested this on two systems, and `nvm install --lts` always uses the latest, making `nvm use --lts` look a bit redundant)

nate-double-u (approved):
New instructions and scripts tested on Ubuntu & macOS.

Other than the one question I have inline, this LGTM

README.md
chalin: I vaguely recall that there were circumstances under which the `use` was needed, but according to https://github.com/nvm-sh/nvm/issues/2057, this is no longer the case since 2014. I'll remove the use of `use`. Nice catch!

chalin: Done

chalin:
Also added a default value for DEPLOY_PRIME_URL, when un set.

README.md
chalin: Ah, now a remember: I was wanting to cater to the scenario that a developer already had Node.js installed -- and so might skip the install command. But, since 2014, I see that it's ok to execute `install` whether or not the latest LTS is already installed. In either case, it is equivalent to following that command with `nvm use`.

PR #427, opened 2021-08-04
Update production-users.md

cheeming:
Updated etcd usage in Grab

### DIFF
--- a/content/en/docs/v3.3/production-users.md
+++ b/content/en/docs/v3.3/production-users.md
@@ -170,16 +170,15 @@ In [hyper.sh][hyper.sh], the container service is backed by [hypernetes][hyperne
 [shafreeck]:https://github.com/shafreeck
 
 ## Grab
-- *Application*: system configuration for services, service discovery
+- *Application*: primarily for service discovery
 - *Launched*: June 2016
 - *Cluster Size*: 1 cluster of 7 members
-- *Order of Data Size*: megabytes
-- *Operator*: Grab, [taxitan][taxitan], [reterVision][reterVision]
+- *Order of Data Size*: hundreds of megabytes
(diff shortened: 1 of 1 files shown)


content/en/docs/v3.3/production-users.md
```diff
@@ -170,16 +170,15 @@ In [hyper.sh][hyper.sh], the container service is backed by [hypernetes][hyperne
...
 
 ## Grab
-- *Application*: system configuration for services, service discovery
+- *Application*: primarily for service discovery
 - *Launched*: June 2016
 - *Cluster Size*: 1 cluster of 7 members
-- *Order of Data Size*: megabytes
-- *Operator*: Grab, [taxitan][taxitan], [reterVision][reterVision]
+- *Order of Data Size*: hundreds of megabytes
+- *Operator*: [Grab]Grab
```
nate-double-u, line 177:
```suggestion
- *Operator*: [Grab][]
```

nate-double-u (changes requested):
Thanks for this @cheeming!

Deploy preview: https://deploy-preview-427--etcd.netlify.app/docs/v3.3/production-users/#grab

As v3.3 is an archived version, it may be better to find an archived version of the links being removed that reflect the links as the were at the time this page was initially published (if they are different now then they were) and replace them with the archive links rather than remove them. The [Wayback Machine](https://web.archive.org/) is often a good tool for this. 

If the pages linked aren't different now than they were when v3.3 was released, then I'm not sure why we would change or remove those links on an archived page.

If we do decide to add a new link, I've suggested a change to the format of the link you provided for Grab in line.

cheeming:
created new one here, https://github.com/etcd-io/website/pull/429

PR #428, opened 2021-08-05
Add tc cancel command.

horizonzy:


nate-double-u:
Hi @horizonzy, thanks for this.

We don't have a style guide for [etcd.io](https://etcd.io), but Kubernetes has a good one that I like to refer to. They suggest [avoiding words that assume a specific level of understanding](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-words-that-assume-a-specific-level-of-understanding), such as "just", "simply", "easy", "easily", or "simple".

In the case of this PR, I'd suggest revising the line "If user want to cancel tc, just execute:" to remove "just" in each file.

nate-double-u:
Sorry, I should have read more closely the first time through:
"If the user wants to cancel `tc`, execute:"

We may want to simplify it further:
"To cancel `tc`, execute:"

nate-double-u (approved):
/lgtm

PR #429, opened 2021-08-09
Update details about etcd usage in Grab

cheeming:
Updated etcd usage in Grab
 - removed users who are no longer actively contributing as etcd operators in Grab

### DIFF
--- a/content/en/docs/v3.1/production-users.md
+++ b/content/en/docs/v3.1/production-users.md
@@ -142,13 +142,12 @@ In [hyper.sh][hyper.sh], the container service is backed by [hypernetes][hyperne
 - *Application*: system configuration for services, service discovery
 - *Launched*: June 2016
 - *Cluster Size*: 1 cluster of 7 members
-- *Order of Data Size*: megabytes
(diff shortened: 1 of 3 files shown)


nate-double-u:
Thanks for this @cheeming!

Non current versions (such as v3.1, v3.2, and v3.3) are provided as archival references for etcd, and we are hesitant to change them with out strong rationale.

As I mentioned in the [previous PR](https://github.com/etcd-io/website/pull/427#pullrequestreview-723583959), it may be better to find an archived version of the sites being removed that reflect the links as they were at the time this page was initially published (if they are different now then they were) and replace these links with the archive links rather than remove them altogether. The [Wayback Machine](https://web.archive.org/) is often a good tool for this. 

If the pages linked aren't different now than they were when these versions were released, then I'm not sure why we would change or remove those links on an archived page.

nate-double-u (changes requested):
Please see [my previous comment](https://github.com/etcd-io/website/pull/429#issuecomment-899708730).

jberkus:
Closing this PR due to lack of updates.

PR #430, opened 2021-08-09
Drop obsolete check-hugo script

chalin:
Followup to #424


nate-double-u (approved):
/lgtm
(tested on Ubuntu)

PR #435, opened 2021-08-13
Add doc for --initial-election-tick-advance flag

### LINKED ISSUE #433: Add doc for --initial-election-tick-advance flag
Originally opened in the etcd repo - https://github.com/etcd-io/etcd/issues/12843 by @pikrzysztof 

It's mentioned in the changelog but I can't find it on https://etcd.io/docs/next/op-guide/configuration/ .

Aman-Codes:
Added doc for `--initial-election-tick-advance` flag
Closes #433

### DIFF
--- a/content/en/docs/v3.1/op-guide/configuration.md
+++ b/content/en/docs/v3.1/op-guide/configuration.md
@@ -43,6 +43,11 @@ To start etcd automatically using custom settings at startup in Linux, using a [
 + default: "1000"
 + env variable: ETCD_ELECTION_TIMEOUT
 
+### --initial-election-tick-advance
++ Whether to fast-forward initial election ticks on boot for faster election. When it is true, then local member fast-forwards election ticks to speed up "initial" leader election trigger. This benefits the case of larger election ticks. Disabling this would slow down initial bootstrap process for cross datacenter deployments. Make your own tradeoffs by configuring this flag at the cost of slow initial bootstrap.
++ default: true
++ env variable: ETCD_INITIAL_ELECTION_TICK_ADVANCE
+
 ### --listen-peer-urls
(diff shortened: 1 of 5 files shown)


content/en/docs/v3.1/op-guide/configuration.md
```diff
@@ -43,6 +43,11 @@ To start etcd automatically using custom settings at startup in Linux, using a [
 + default: "1000"
 + env variable: ETCD_ELECTION_TIMEOUT
 
+### --initial-election-tick-advance
++ Whether to fast-forward initial election ticks on boot for faster election. When it is true, then local member fast-forwards election ticks to speed up "initial" leader election trigger. This benefits the case of larger election ticks. Disabling this would slow down initial bootstrap process for cross datacenter deployments. Make your own tradeoffs by configuring this flag at the cost of slow initial bootstrap.
++ default: true
++ env variable: ETCD_INITIAL_ELECTION_TICK_ADVANCE
```
chalin, line 49: It might be worth adding a bullet stating that this was added in 3.1.14, and link to the corresponding change-log entry: https://github.com/etcd-io/etcd/blob/main/CHANGELOG-3.1.md#v3114-2018-04-24

The note would only be needed here in the v3.1 docs, not the docs for > v3.1.

@spzala - what do you think?

chalin (commented):
Overall, LGTM, though I'll let @spzala comment as well. Also see inlined comment.

content/en/docs/v3.1/op-guide/configuration.md
spzala: @chalin thanks for your review! I will provide my comments today.

spzala: Good point but I think it's OK not to add it. Seems like the flag was added during 3.4 development and backported. Usually we don't add when the flag was added explicitly here as in rest of the doc (added explicitly in the changelog only). Thanks!

spzala (approved):
lgtm
Thanks @Aman-Codes

PR #439, opened 2021-08-13
Consolidate CI and regular link-check commands

chalin:


chalin:
More changes are coming in. I'll merge this now.

PR #446, opened 2021-08-16
Remove the outdated gif screenshot and refine the behavior of auth

mitake:
The outdated documentation related to auth was reported in https://github.com/etcd-io/etcd/issues/12966
This PR fixes the issue. It removes the screenshot gif and enhance the command line description.

The behavior of the root role was fixed in https://github.com/etcd-io/etcd/issues/6355 , the root can access to every key so it doesn't cause permission errors described in the current doc.

### DIFF
--- a/content/en/docs/v3.1/demo.md
+++ b/content/en/docs/v3.1/demo.md
@@ -432,25 +432,30 @@ etcd --data-dir=data.etcd --name ${THIS_NAME} \
 
 `auth`,`user`,`role` for authentication:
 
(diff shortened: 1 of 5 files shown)


mitake:
@gyuho if you have recommendation for gif screenshot (font and size for terminal emulator, etc), I'll take a new one. But it won't be consistent with the existing screenshots so I just deleted the outdated one.

spzala (approved):
LGTM
Thanks @mitake

PR #451, opened 2021-08-18
fix v2.3 links

kapunahelewong:
Updates links as per https://github.com/etcd-io/website/issues/419

cc @chalin


chalin (approved):
Ah, a nice short section! LGTM ✨

PR #454, opened 2021-08-19
fix v3.2 links

kapunahelewong:
Fixes links in `v3.2` links as per https://github.com/etcd-io/website/issues/419.

cc @chalin


chalin (approved):
LGTM, thanks! (Probably starting to feel like déjà vu 😄)

Issue #245, opened 2021-04-26
Remaining broken external links


nate-double-u:
I've gotten this to a point where there are a handful broken links left, and I'm not sure where they should point:

docs/v3.4/learning/why/index.html
* Non-OK status: 404 --- docs/v3.4/learning/why/index.html --> https://godoc.org/github.com/coreos/etcd-io/etcdserver/api/v3election/v3electionpb

docs/v3.3/learning/why/index.html
* Non-OK status: 404 --- docs/v3.3/learning/why/index.html --> https://godoc.org/github.com/coreos/etcd-io/etcdserver/api/v3election/v3electionpb

docs/v3.1/libraries-and-tools/index.html
* Non-OK status: 404 --- docs/v3.1/libraries-and-tools/index.html --> https://github.com/scrz/scrz

docs/v3.1/op-guide/monitoring/index.html
* request exceeded our ExternalTimeout --- docs/v3.1/op-guide/monitoring/index.html --> http://dash.etcd.io/dashboard/db/test-etcd

docs/next/learning/why/index.html
* Non-OK status: 404 --- docs/next/learning/why/index.html --> https://godoc.org/github.com/etcd-io/etcd/server/etcdserver/api/v3lock/v3lockpb
* Non-OK status: 404 --- docs/next/learning/why/index.html --> https://godoc.org/github.com/etcd-io/etcd/server/etcdserver/api/v3election/v3electionpb
* Non-OK status: 404 --- docs/next/learning/why/index.html --> https://godoc.org/github.com/etcd-io/etcd/server/etcdserver/api/v3lock/v3lockpb

At this point, I think we should review this PR, and open a separate issue for the remaining links.

_Originally posted by @nate-double-u in https://github.com/etcd-io/website/issues/203#issuecomment-823642654_

nate-double-u:
`make check-all-links` output:

```
htmltest started at 09:25:45 on public
========================================================================
blog/jepsen-343-results/index.html
  Non-OK status: 416 --- blog/jepsen-343-results/index.html --> https://fauna.com/blog/demystifying-database-systems-part-4-isolation-levels-vs-consistency-levels
docs/v3.4/learning/why/index.html
  Non-OK status: 404 --- docs/v3.4/learning/why/index.html --> https://godoc.org/github.com/coreos/etcd-io/etcdserver/api/v3election/v3electionpb
  Non-OK status: 416 --- docs/v3.4/learning/why/index.html --> https://www.consul.io/commands/lock
[...]

PR #456, opened 2021-08-19
Streamline link-checking build config and commands

chalin:
- Use `--skip-external` flag instead of having two `htmltest` config files.
- As necessary, build `htmltest` from source so that we have the latest features!
  - Once build,`htmltest` will be reused (rather than refetched & rebuilt).
- Ignore `/docs/latest` so that we no longer need to setup symlinks

---

If you've previously run the link checker scripts, you'll first need to delete the `HTMLTEST` directory before running the link checking scripts again. To do this run this command:

```sh
# From your etcd.io website repo directory
$ rm -Rf tmp
```

### DIFF
--- a/.htmltest.external.yml
+++ b/.htmltest.external.yml
@@ -1,4 +1,0 @@
-DirectoryPath: tmp/public
-IgnoreDirectoryMissingTrailingSlash: true
-IgnoreAltMissing: true
-IgnoreURLs: [example.com]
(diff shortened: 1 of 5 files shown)


nate-double-u (approved):
/lgtm

Tested on Ubuntu.

The `npm run check-links:all` throws errors that it probably shouldn't but I think that's outside of the scope of this PR, and reflects that we likely need to revisit external link checking as a separate issue.

chalin:
> we likely need to revisit external link checking as a separate issue.

Right. I also mentioned that in https://github.com/etcd-io/website/issues/245#issuecomment-902226903, but maybe we need to open a separate issue.

PR #457, opened 2021-08-19
fix v3.5 links

kapunahelewong:
Fixes v3.5 links as per https://github.com/etcd-io/website/issues/419

cc @chalin


chalin (approved):
Wonderful! We're getting closer to our final goal! ✨

PR #458, opened 2021-08-20
fix 3.4 links

kapunahelewong:
Fix links in v3.4 per https://github.com/etcd-io/website/issues/419

cc @chalin


chalin (approved):
I added a couple of missing trailing slashes, but otherwise, well done!

PR #460, opened 2021-08-23
README: warn about temporary extra setup before links can be checked

chalin:


README.md
```diff
@@ -29,6 +29,10 @@ To locally serve the site at [localhost:8888][], run the following command:
 $ npm run serve
 ```
 
+> **WARNING:** If you want to **check links** locally, you (temporarily) might have
```
nate-double-u, line 32: I'm not sure this should be a warning -- the build won't break if the link checker doesn't run locally. Maybe `Note`, or `Caution`?

chalin: Good point. I went with **NOTE**.

nate-double-u (approved):

PR #468, opened 2021-09-03
Converted the page Platforms/Amazon Web Services into blog post

Somoshree:
closes: #396

PR #469, opened 2021-09-05
Converted Platforms/Container-Linux-with-systemd page into blog post …

Somoshree:
closes: #397


nate-double-u:
Deploy preview: https://deploy-preview-469--etcd.netlify.app/blog/2021/etcd-container-linux-with-systemd/
Redirect test: https://deploy-preview-469--etcd.netlify.app/docs/v3.5/platforms/container-linux-systemd/ ✔️

nate-double-u:
Here is the history of the original page from the etcd-io/etcd repository:
https://github.com/etcd-io/etcd/commits/8b290c680a5d4b4067c4d91b33f70c5506f654f0/Documentation/platforms/container-linux-systemd.md

I believe @gyuho was the original author of that page, though I think we may need to investigate a bit more as the file name is different.

Okeanos:
This article should probably be backdated way, way into the past and/or immediately marked as deprecated. Container Linux as such doesn't exist anymore and the [successor Fedora CoreOS](https://docs.fedoraproject.org/en-US/fedora-coreos/faq/#_does_fedora_coreos_replace_container_linux_what_happens_to_cl) is different enough that the contents of this article don't make much sense.

It is probably easier to refer to the [Fedora CoreOS documentation on etcd](https://docs.fedoraproject.org/en-US/fedora-coreos/running-containers/#running-etcd) which explains the bare necessities of the Fedora CoreOS runtime and also links to the official documentation on running etcd in a container.

nate-double-u:
@Okeanos, back dating is a good idea. I think we're running into the same issue over on https://github.com/etcd-io/website/pull/468#discussion_r714312926 as well. Should we date for just after the original page was built?, when it was moved?

@Somoshree, @Okeanos makes a good point about the content being old, but maybe we could write a caveat pointing people in the direction of [Fedora CoreOS documentation on etcd](https://docs.fedoraproject.org/en-US/fedora-coreos/running-containers/#running-etcd), and that this post is provided as an archival reference?

content/en/blog/2021/etcd-container-linux-with-systemd.md
```diff
@@ -1,10 +1,11 @@
 ---
 title: Container Linux with systemd
-weight: 7200
+author: Gyuho Lee, Somoshree Datta
```
spzala, line 3: Just my thought but I would only keep Gyuho's name here, considering the technical content, if we really need author.  Also, same comment with date as my previous comment in a related PR. Thanks @Somoshree @nate-double-u

spzala (changes requested):

content/en/blog/2021/etcd-container-linux-with-systemd.md
nate-double-u: In https://github.com/etcd-io/website/pull/468 we've added a note at the bottom of the page explaining the provenance of the page. I think we can do that here too, remove the author line, and add a similar note here.

spzala (approved):
lgmt
thanks @Somoshree @nate-double-u

Issue #394, opened 2021-06-24
Write "Getting Started with `etcdctl`" page


nate-double-u:
As per #65:
> Getting Started with Etcdctl - page with simple instructions on how to install and get started with Etcdctl, including authentication to existing clusters.

This page could be linked to from both the Quickstart and Install pages, but should stand on its own content wise.

nate-double-u:
https://etcd.io/docs/v3.5/dev-guide/interacting_v3/ may be a good base/starting point for this page.

spzala:
https://github.com/etcd-io/etcd/tree/main/etcdutl

nate-double-u:
Thanks for this @Somoshree, it's a great start!

I see that you're basing this page on the [Interacting with etcd page](https://etcd.io/docs/v3.5/dev-guide/interacting_v3/), I think that's a good idea, but remember the goals of that page and a getting started page are different.

A getting started page should show a reader enough to get a simple example system up and running, and link out to other pages for more detail or advanced setup.

In our case, etcd.io may not have the content for these advanced detail links - if that's the case, it's OK to link to the etcd-io/etcd repo that @spzala shared. If we do go that route, we'll want to open another issue to ensure that we eventually bring the https://github.com/etcd-io/etcd/tree/main/etcdutl content into the main documentation repo (here), but doing any more than linking to it  in this PR is probably out of scope for creating a getting started page.

Also note, if you're taking content out of the [Interacting with etcd page](https://etcd.io/docs/v3.5/dev-guide/interacting_v3/) to bring it into the getting started page, we'll want to make sure it's removed from that page so the content isn't duplicated. That page may be need to be edited and reworked a bit in the context of having a new getting started guide.

Here are some Getting Started pages that I think are good examples:
* https://falco.org/docs/getting-started/
* https://kubernetes.io/docs/setup/
* https://etcd.io/docs/v3.5/quickstart/

PR #476, opened 2021-09-13
Added 'Getting started with etcdctl' page and also modified Quickstart page to link to the newly added page

Somoshree:
closes: #394 

Deploy previews:
+ https://deploy-preview-476--etcd.netlify.app/docs/v3.5/getting-started-with-etcdctl/
+ https://deploy-preview-476--etcd.netlify.app/docs/v3.5/quickstart/#whats-next


nate-double-u:
Deploy previews:
* https://deploy-preview-476--etcd.netlify.app/docs/v3.5/getting-started-with-etcdctl/
* https://deploy-preview-476--etcd.netlify.app/docs/v3.5/quickstart/#whats-next

nate-double-u:
Thanks for this @Somoshree, it's a great start!

I see that you're basing this page on the [Interacting with etcd page](https://etcd.io/docs/v3.5/dev-guide/interacting_v3/), I think that's a good idea, but remember the goals of that page and a getting started page are different.

A getting started page should show a reader enough to get a simple example system up and running, and link out to other pages for more detail or advanced setup.

In our case, etcd.io may not have the content for these advanced detail links - if that's the case, it's OK to link to the etcd-io/etcd repo that @spzala shared. If we do go that route, we'll want to open another issue to ensure that we eventually bring the https://github.com/etcd-io/etcd/tree/main/etcdutl content into the main documentation repo (here), but doing any more than linking to it  in this PR is probably out of scope for creating a getting started page.

Also note, if you're taking content out of the [Interacting with etcd page](https://etcd.io/docs/v3.5/dev-guide/interacting_v3/) to bring it into the getting started page, we'll want to make sure it's removed from that page so the content isn't duplicated. That page may be need to be edited and reworked a bit in the context of having a new getting started guide.

Here are some Getting Started pages that I think are good examples:
* https://falco.org/docs/getting-started/
* https://kubernetes.io/docs/setup/
* https://etcd.io/docs/v3.5/quickstart/

nate-double-u:
These are some good changes @Somoshree. The shorter page with more direct instructions is good, but be careful that we're not duplicating content from other pages directly.

The other thing to look at, from the issue, is:
> Getting Started with Etcdctl ... including authentication to existing clusters.

@spzala is there a good source to reference for authenticating to existing clusters with `etcdctl`? Somewhere on http://play.etcd.io perhaps?

Also, it may be difficult to include authentication to existing clusters unless we in the earlier instructions set up a cluster to authenticate to, or perhaps we can set up a docker container somewhere to help stream line the process.

spzala (changes requested):
Suggest to remove lines from 8 to 16. We don't need them here. 
Also, remove lines 17 to 23, and add something like "etcdctl is shipped with etcd release binaries. On a running etcd cluster, find out the version of etcdctl by running the following command..." Thanks!

spzala:
@nate-double-u here, https://github.com/etcd-io/etcd/tree/main/etcdctl#auth-enable-or-disable 
I think we should reference "https://github.com/etcd-io/etcd/tree/main/etcdctl" somewhere in the interacting with etcd page.. The "https://github.com/etcd-io/etcd/tree/main/etcdctl" will be more latest with any code changes.

nate-double-u:
In a meeting earlier today with @Somoshree & @spzala, we discussed potentially updating the [Quickstart](https://etcd.io/docs/v3.5/quickstart/) page with information about _authentication to existing clusters_, rather than creating a new page here as a lot of the content is a duplication of sorts. 

Potentially we don't need a separate "Getting Started with `etcdctl`" page when the Quickstart page does so much with `etcdctl` already.

nate-double-u:
Closing as per conversation with @Somoshree & @jberkus

PR #493, opened 2021-09-20
Style fix of offline-search popup close button, w/o a docsy override

chalin:
Followup to #403

### Screenshot

> image

Preview: https://deploy-preview-493--etcd.netlify.app/


nate-double-u (approved):
This is better, thanks @chalin

Issue #494, opened 2021-09-20
Upgrade docsy


chalin:
- [x] #431
- [ ] Upgrade docsy proper
- [ ] Revert the CSS change of #493 since it is addressed in https://github.com/google/docsy/pull/500

PR #497, opened 2021-09-21
Improved sidebar indentation

chalin:
- Followup to #494
- Removes sidebar indentation from version top-level pages, giving us more sidebar real estate for nested pages. (We don't lose anything w/o that level of indentation because one is never looking at the pages across two versions at the same time, and it is clear which version section we're in.)
- Use the same sidebar indentation for desktop & mobile (as opposed to the default mobile all-flat sidebar)

### Desktop

Before:

> image

After:

> image

### Mobile

Before:

> 

After:

>

### DIFF
--- a/assets/scss/_styles_project.scss
+++ b/assets/scss/_styles_project.scss
@@ -158,6 +158,11 @@
   padding-top: 4.75rem;
 }
 
+.td-sidebar-nav__section {
+  .ul-1 ul { padding-left: 1rem; }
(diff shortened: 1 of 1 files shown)


nate-double-u (approved):
/lgtm 

tested on:
* Firefox 92 on Ubuntu 20.04.1
* Safari on iOS 14.8
* Safari 14.1.2 on macOS 10.15.7

PR #519, opened 2021-10-11
Fix v3.5 import package path error in gRPC naming document

cyub:
Fix v3.5 import package path error in gRPC naming document


jberkus:
Apologies for leaving this open; the same fix was merged from someone else.  Much thanks for offering the correction.

PR #522, opened 2021-10-15
Document the user experience change since v3.5.0 on the `/health` endpoint when authentication is enabled.

ahrtr:
fix https://github.com/etcd-io/etcd/issues/13415 . 

Previously in etcd 3.4, the requests on `/health` are not authenticated even the authentication is enabled. But since etcd v3.5.0, it will be authenticated if the authentication is enabled. So we need to document the user experience change in 3.5.0.


ahrtr:
It should have already been fixed in 3.5.1.  See [this](https://github.com/etcd-io/etcd/commit/dfd2fea4c5d259ebab4e15b05aee054a6c528394)

PR #539, opened 2021-11-11
Adding cloud IDE (https://gitpod.io) config file and instructions

nate-double-u:
I've been working with gitpod.io cloud development environments, and have found it very useful, so I thought I'd make it a  part of this repo.

This change adds a configuration file and instructions for gitpod. When you visit https://gitpod.io/#https://github.com/etcd-io/website, a Gitpod.io IDE is launched that allows you to build, preview and make changes to this repo.

_Note_, visiting the link provided doesn't launch the preview automatically yet as it grabs the `main` branch (it does bring up the IDE, and you can run the preview manually). You can see how it would work by looking at the [Notary Project's](https://github.com/notaryproject/notaryproject.dev) setup: https://gitpod.io/#https://github.com/notaryproject/notaryproject.dev

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,11 @@
+# Commands to start on workspace startup
+tasks:
+  - name: Dev Server
+    init: npm install
+    command: npm run serve
+
+# Ports to expose on workspace startup
+ports:
+  - port: 8888
+    onOpen: open-browser
+
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,7 +1,9 @@
 # Commands to start on workspace startup
 tasks:
   - name: Dev Server
-    init: npm install
+    init: |
+      nvm install --lts
+      npm install
     command: npm run serve
 
 # Ports to expose on workspace startup
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,7 +1,7 @@
 # Commands to start on workspace startup
 tasks:
   - name: Dev Server
-    init: npm install
+    init: npm install --lts
     command: npm run serve
 
 # Ports to expose on workspace startup
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,7 +1,7 @@
 # Commands to start on workspace startup
 tasks:
   - name: Dev Server
-    init: npm install --lts
+    init: npm install
     command: npm run serve
 
 # Ports to expose on workspace startup
--- a/README.md
+++ b/README.md
@@ -2,13 +2,11 @@
 
 The [etcd.io][] website, built using [Hugo][] and hosted on [Netlify][].
 
-## Build prerequisites
-
-### Cloud build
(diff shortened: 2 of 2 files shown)


.gitpod.yml
```diff
@@ -0,0 +1,11 @@
+# Commands to start on workspace startup
+tasks:
+  - name: Dev Server
+    init: npm install
```
nate-double-u, line 4: @chalin, should this be:

```suggestion
    init: npm install --lts
```

nate-double-u:
@Somoshree this may be interesting to you as well.

.gitpod.yml
chalin: Yes

README.md
```diff
@@ -4,6 +4,12 @@ The [etcd.io][] website, built using [Hugo][] and hosted on [Netlify][].
 
 ## Build prerequisites
 
+### Cloud build
+
+Visit [https://gitpod.io/#https://github.com/etcd-io/website](https://gitpod.io/#https://github.com/etcd-io/website) to launch a [Gitpod.io](https://gitpod.io) IDE that will allow you to build, preview and make changes to this repo.
```
chalin, line 9: The cloud build doesn't have any "Build prerequisites" so it should be, for example, the first second-level heading in this page.

README.md
```diff
@@ -4,6 +4,12 @@ The [etcd.io][] website, built using [Hugo][] and hosted on [Netlify][].
...
+
+### Local build
+
```
chalin, line 12: IMHO, you can drop this addition and instead edit the level-2 heading to read:

> ## Local build

If you do so, then avoid the subheading promotions below.

chalin (commented):
See inline comments.

nate-double-u:
Readme file deploy preview: https://github.com/etcd-io/website/blob/add-cloud-ide/README.md

.gitpod.yml
```diff
@@ -0,0 +1,11 @@
+# Commands to start on workspace startup
+tasks:
+  - name: Dev Server
+    init: npm install --lts
```
chalin, line 4: Oh, sorry, I wasn't careful when I commented on this line before. The `--lts` is needed for `nvm install` **not** `npm install`.

So this line should read: `init: npm install`

chalin (commented):
One change to revert (my bad). Otherwise LGTM

.gitpod.yml
chalin: If somehow we can specify Node constraints here, then we want the latest LTS version.

nate-double-u: thanks for the clarification! I'll look into it, we may be able to have both run.

nate-double-u: I've updated the .gitpod.yml, it looks like we can run  multiple commands using `|`

chalin (approved):

Issue #538, opened 2021-11-09
Docs, Blog, and Install navbar entries are not shown as active when pages are being viewed


nate-double-u:
- [ ] Docs
- [x] Blog (see screenshot below) - #542
- [ ] Install

related: https://github.com/grpc/grpc.io/issues/895

### expected
<img width="1169" alt="Screen Shot 2021-11-09 at 3 21 10 PM" src="https://user-images.githubusercontent.com/4453979/141021406-7f4f66d7-6e6f-4752-821e-5a5543714ba3.png">

### actual
<img width="1169" alt="Screen Shot 2021-11-09 at 3 21 22 PM" src="https://user-images.githubusercontent.com/4453979/141021405-1d528ff1-2fe9-4352-992a-e71e0ed83bb3.png">

nate-double-u:
Marking as "good first issue" as the [related issue](https://github.com/grpc/grpc.io/issues/895) has 2 PRs that show potential solutions.

nate-double-u:
Another potential solution: https://github.com/gohugoio/hugo/issues/9150#issuecomment-965798975

nate-double-u:
Closing as duplicate of https://github.com/etcd-io/website/issues/443

PR #542, opened 2021-11-18
Ensure Blog top-nav entry is marked as active for blog pages

chalin:
Contributes to #538, fixing it for blog pages. The other pages are trickier because they are not real links. I'll have to think about those cases a bit more.

Preview: https://deploy-preview-542--etcd.netlify.app/blog

### Screenshot

image

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -169,7 +169,7 @@ menu:
       url: /docs/latest/
       weight: -10
     - name: Blog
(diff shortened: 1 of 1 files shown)


chalin:
Going to merge this now so that I can tackle the other sections.

nate-double-u:
Thanks @chalin!

PR #548, opened 2021-12-09
Update monitoring.md

eternal-dragon:
https://github.com/etcd-io/website/blob/main/content/en/docs/v3.5/upgrades/upgrade_3_5.md#deprecated-etcd---debug-flag-now---log-leveldebug

### DIFF
--- a/content/en/docs/v3.5/op-guide/monitoring.md
+++ b/content/en/docs/v3.5/op-guide/monitoring.md
@@ -8,7 +8,7 @@ Each etcd server provides local monitoring information on its client port throug
 
 ## Debug endpoint
 
-If `--debug` is set, the etcd server exports debugging information on its client port under the `/debug` path. Take care when setting `--debug`, since there will be degraded performance and verbose logging.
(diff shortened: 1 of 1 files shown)


content/en/docs/v3.5/op-guide/monitoring.md
```diff
@@ -8,7 +8,7 @@ Each etcd server provides local monitoring information on its client port throug
 
 ## Debug endpoint
 
-If `--debug` is set, the etcd server exports debugging information on its client port under the `/debug` path. Take care when setting `--debug`, since there will be degraded performance and verbose logging.
+If `--log-level=debug` is set, the etcd server exports debugging information on its client port under the `/debug` path. Take care when setting `--log-level=debug`, since there will be degraded performance and verbose logging.
```
spzala, line 11: Thanks @eternal-dragon This is fine but you should also update the `--debug`` reference here as well https://github.com/etcd-io/website/blob/main/content/en/docs/v3.5/faq.md

jberkus:
This is still a correct fix regardless of whether we need to fix it elsewhere; merging.

PR #550, opened 2021-12-27
Update supported-platform.md

jiangxiaobin96:


jmhbnz:
Hi @jiangxiaobin96, thanks for raising this suggestion. Apologies you did not receive a timely response to this pull request, the etcd project has not had enough contributors to respond to all issues.

I am closing this pull request as from etcd [issues and pull requests](https://github.com/etcd-io/etcd/issues?q=risc) I don't believe support for risc was ever actually added and I don't think it is a priority for the project at the moment.

We can always revisit this in future and I would suggest a starting point would be adding some kind of continuous integration support for risc in the etcd-io/etcd repo.

PR #552, opened 2022-01-12
Added OS package installation instructions

jdato:


jmhbnz (approved):
LGTM

@ahrtr , @spzala - This is a quick improvement to our install docs, can you please take a look?

It's been a long time since opening so would now need to be replicated to 3.6 alpha docs which I don't mind doing in a followup.

jmhbnz:
Ping @ahrtr, @serathius, @spzala - This fills a gap in our install docs can you please take a look?  I don't mind doing the followup to replicate to 3.6.

serathius:
Don't know about Mac, but debian packages are not maintained and very outdated https://packages.debian.org/bullseye/etcd. Last release is 3 years old https://github.com/etcd-io/etcd/releases/tag/v3.3.25

jmhbnz:
> Don't know about Mac, but debian packages are not maintained and very outdated https://packages.debian.org/bullseye/etcd. Last release is 3 years old https://github.com/etcd-io/etcd/releases/tag/v3.3.25

Good spotting on this @serathius.  In that case we either remove the Debian section, or we can add a disclaimer. I am ok with either approach.

jdato:
I just tested with `apt`, `apt-get` and `aptitude` and all of them install the older version 3.3.25. I've added a disclaimer to the section.

content/en/docs/v3.5/install.md
```diff
@@ -71,8 +71,49 @@ source by following these steps:
...
+```sh
+$ brew install etcd
+```
+
+3. Verify install 
+```sh
+$ etcd --version
+```
+
+### Debian Linux (apt)
```
serathius, line 93: I would not want to recommend to users unsupported and unmaintained way to install etcd. Let's drop it or improve the situation.

ahrtr:
I am not sure how much additional effort is needed to maintain Debian/Ubuntu etcd packages.  I would suggest to mark it as TBD, and update it until the package is up to date and the maintainers & mejbers are agree to maintain it.

Regarding to the Mac etcd package, I guess it's automatically maintained? Could anyone clearly clarify it?

content/en/docs/v3.5/install.md
jmhbnz: Just for everyone's info there has been some discussion earlier this year on getting an updated etcd package in but it looks like it has stalled: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=991462 

We can remove it for now sure. I'll keep an eye on it and if the situation improves I can raise a pr.

jdato:
> Regarding to the Mac etcd package, I guess it's automatically maintained? Could anyone clearly clarify it?

As far as I understand it `homebrew` is also not updated automatically, hence potentially bears the same issues as `apt` sometime in the future. Maybe it makes more sense to just keep the installation instructions for specific OS out of the docs for now. And perhaps just add a note in that section that there are ways to install through package managers, however unsupported by etcd project.

jmhbnz:
> I am not sure how much additional effort is needed to maintain Debian/Ubuntu etcd packages. I would suggest to mark it as TBD, and update it until the package is up to date and the maintainers & mejbers are agree to maintain it.
> 
> Regarding to the Mac etcd package, I guess it's automatically maintained? Could anyone clearly clarify it?

Here is the homebrew etcd page: https://formulae.brew.sh/formula/etcd

And here is the backing ruby code.  Reviewing the blame the tags are not automatically updated but it does appear to be regularly maintained by individuals so it makes sense for me for us to offer that method to users on the caveat it isn't officially supported by the project: https://github.com/Homebrew/homebrew-core/blob/06a368c0355ff1e49a454896a6face806733675d/Formula/etcd.rb

jdato:
I've now removed the outdated debian section, added a more generic disclaimer with note that etcd project doesn't support OS package manager installs and kept the homebrew install instructions, since it seems to be kept updated.

jmhbnz:
> I've now removed the outdated debian section, added a more generic disclaimer with note that etcd project doesn't support OS package manager installs and kept the homebrew install instructions, since it seems to be kept updated.

Thanks @jdato.   Can you please squash commits i.e;

```bash
git reset --soft HEAD~5
git commit -s -a -m "Added OS package installation instructions"
git push --force
```

Once squashed in my view the changes are ready to merge 👍🏻

PR #554, opened 2022-01-14
fix golang example of with a lease

oofpgDLD:
fix golang example of registering an endpoint with a lease


spzala (approved):
lgtm
Thanks @oofpgDLD

PR #555, opened 2022-01-20
Link to etcd-postgresql-sync and etcd-client-from-env

Jille:
The first is a daemon that syncs etcd's KVs (and authentication config) to PostgreSQL, which is useful for backups with little extra infrastructure.
The latter a library that reads environment variables for connecting to an etcd cluster, on which it would be nice to standardize for user convenience.


jmhbnz (commented):
Hi @Jille - Thanks for raising this suggestion. Apologies that it has taken a long time for someone to look at this pull request, the etcd project has not had enough contributors to handle all request.

If you would still like to have this merged can you please resolve the conflicts.

Additionally, I notice `etcd-client-from-env` has not had any maintenance in two years so please consider if that is still something you want to include in this request. Thanks!

jmhbnz:
Closing due to lack of activity.

PR #556, opened 2022-01-31
Canceled grpc is now not considered an error

ksa-real:
"Canceled" grpc state means client-initiated cancellation of streaming
grpc call, e.g. Watch. This is a legitimate non-error case. Before this
change such codes were considered an error and created false positive
alerts.

### DIFF
--- a/content/en/docs/v3.4/op-guide/etcd3_alert.rules.yml
+++ b/content/en/docs/v3.4/op-guide/etcd3_alert.rules.yml
@@ -34,7 +34,7 @@ groups:
       message: 'etcd cluster "{{ $labels.job }}": {{ $value }}% of requests for {{
         $labels.grpc_method }} failed on etcd instance {{ $labels.instance }}.'
     expr: |
-      100 * sum(rate(grpc_server_handled_total{job=~".*etcd.*", grpc_code!="OK"}[5m])) BY (job, instance, grpc_service, grpc_method)
+      100 * sum(rate(grpc_server_handled_total{job=~".*etcd.*", grpc_code!~"OK|Canceled"}[5m])) BY (job, instance, grpc_service, grpc_method)
         /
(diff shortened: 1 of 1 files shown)


ksa-real:
@spzala Can you please take a look and suggest how to proceed? I see that `op-guide/etcd3_alert.rules.yml` is not present anymore in op-guide in v3.5 and v3.6. The v3.4 was the latest. Are the rules not supported anymore? What is the new place for the source of truth for the rules? The issue is that [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack/hack#sync_prometheus_rulespy) is using etcd-io rules for its setup. At the moment I think it still takes rules from v3.4.

spzala:
@ksa-real so as I mentioned above, the issue as mentioned in this PR is taken care in the, https://github.com/etcd-io/etcd/blob/master/contrib/mixin/mixin.libsonnet
Isn't it that the kube-prometheus use the `mixin.libsonnet` file?. This PR has some related old discussions, https://github.com/etcd-io/etcd/pull/12689

spzala:
@ksa-real yrw! If changing `etcd3_alert.rules.yml`per your PR helps Kube Prometheus Stack, I think that should be fine and we can merge it after review. Does that sounds good? 
About your second point - this was discussed under https://github.com/etcd-io/etcd/pull/13127 and the decision was to go with the current approach, so I think it's good to leave it that way. In the future, if we see this approach needs to be changed due to maintenance or usability reasons, please propose accordingly. 
Also, I have created a new file from https://github.com/etcd-io/etcd/blob/release-3.5/contrib/mixin/mixin.libsonnet for the v3.5, if that sounds good? - https://github.com/etcd-io/website/pull/600

ksa-real:
SGTM. As for the second point, I guess GRPC codes change rarely, so I don't have too high motivation to insist :) I think I'll leave my arguments at the original discussion and close this diff.

spzala:
Ok, cool :) Thanks @ksa-real

PR #557, opened 2022-02-04
Fix typo in data_model.md

gcjensen:


nate-double-u (approved):
Thanks for this @gcjensen

PR #565, opened 2022-02-28
Create doc for v3.6

spzala:
The v3.6.0-alpha has recently released.

This PR creates base for the v3.6 so that docs can be quickly
updated as we progress further with the v3.6.


spzala:
@nate-double-u making this changes to start with. I am keeping the title to alpha for now, but everything else to 3.6 so that when we have 3.6 releases, we just need to update the title. Hope this makes sense. Thanks!

spzala:
cc @serathius @ptabor

serathius (approved):

ahrtr (approved):

spzala:
ping @nate-double-u  :)

content/en/docs/v3.6/_index.md
```diff
@@ -0,0 +1,14 @@
+---
+title: v3.6.0-alpha docs
+cascade:
+  version: v3.6
+  versName: &name v3.6
+  git_version_tag: v3.6.0
+  exclude_search: false
+linkTitle: *name
+simple_list: true
+weight: -350 # Weight for doc version vX.Y should be -XY0
```
nate-double-u, line 10:
```suggestion
weight: -360 # Weight for doc version vX.Y should be -XY0
```
This will sort this version on the top of the left hand menu.

nate-double-u (approved):
I've only made one minor suggestion. Otherwise this looks good. (Sorry for the delay in reviewing)

Deploy preview: https://deploy-preview-565--etcd.netlify.app/

content/en/docs/v3.6/_index.md
spzala: Good catch! Sounds good. Thanks @nate-double-u

spzala:
> I've only made one minor suggestion. Otherwise this looks good. (Sorry for the delay in reviewing)
> 
> Deploy preview: https://deploy-preview-565--etcd.netlify.app/

Great suggestion. No worries on delay, thanks @nate-double-u !!

spzala:
@nate-double-u done, can you please create a preview to make sure it look good? Thanks!

content/en/docs/v3.5/_index.md
```diff
@@ -7,7 +7,7 @@ cascade:
   exclude_search: false
 linkTitle: *name
 simple_list: true
-weight: -350 # Weight for doc version vX.Y should be -XY0
+weight: -360 # Weight for doc version vX.Y should be -XY0
```
nate-double-u, line 10: this one should stay `-350` -- if it's the same as v3.6's  weight, then the ordering on the left hand menu defaults to alphabetical and v3.5 will come in a above v3.6

spzala: @nate-double-u ah, I got mixed up with the files names. Good catch.

nate-double-u:
We may not want to update it yet, as it's still an alpha, but we may want to change the `layouts/index.redirects` file -- things like `/docs/next` should point at `/docs/v3.6/`

spzala:
@nate-double-u I was hopping that we can update it with alpha and just remove the alpha once we go live with 3.6. That will be quick change and less work.  From your preview link, the doc will look like https://deploy-preview-565--etcd.netlify.app/docs/v3.6/ which looks good to me unless you think it's not good and strongly suggest to go other way.

spzala:
@nate-double-u ah, if the Next is going to v3.5 then yes I totally agree. I thought with these changes, the next will point to 3.6 alpha. So sounds good, I will work on redirect before merging this PR.

spzala:
@nate-double-u updated the `layouts/index.redirects` file. Thanks!

config.yaml
```diff
@@ -82,8 +82,9 @@ params:
   # This menu appears only if you have at least one [params.versions] set.
   version_menu: Versions
   versions:
-    latest: v3.5
+    latest: v3.6
```
nate-double-u, line 85: I think `latest` should stay `v3.5` until `v3.6` comes out of alpha, and is actually released. I'd suggest updating this the day of the release.

spzala: Ok, sound good. Thanks @nate-double-u

PR #572, opened 2022-03-26
Rectify build script path

dengziming:
In etcd source code repo, build.sh is in /scripts instead of ./


spzala (changes requested):
@dengziming this seems not changed in v3.5, but it's needed for v3.6. If you agree, it would be good if you can make this change in v3.6. Thanks!

spzala:
Closing as fixed - https://etcd.io/docs/v3.6/install/. Thanks!

PR #582, opened 2022-05-04
Modify the old url 

my-git9:
The url `http://kubernetes.io/docs/whatisk8s` has been changed to `https://kubernetes.io/docs/concepts/overview/what-is-kubernetes/`,
I think we shoud to modify it

/kind cleanup


spzala (approved):
Thanks @nate-double-u

PR #583, opened 2022-05-07
Update the discovery service protocol for v3 discovery

ahrtr:
The related issue is [issues/13624](https://github.com/etcd-io/etcd/issues/13624), and PR is [pull/13635](https://github.com/etcd-io/etcd/pull/13635)

cc @philips @lucperkins @serathius @ptabor @spzala

### DIFF
--- a/content/en/docs/v3.6/dev-internal/discovery_protocol.md
+++ b/content/en/docs/v3.6/dev-internal/discovery_protocol.md
@@ -4,21 +4,24 @@ weight: 1500
 description: Discover other etcd members in a cluster bootstrap phase
 ---
 
-Discovery service protocol helps new etcd member to discover all other members in cluster bootstrap phase using a shared discovery URL.
+Discovery service protocol helps new etcd member to discover all other members in cluster bootstrap phase using a shared discovery token and endpoint list.
 
 Discovery service protocol is _only_ used in cluster bootstrap phase, and cannot be used for runtime reconfiguration or cluster monitoring.
 
 The protocol uses a new discovery token to bootstrap one _unique_ etcd cluster. Remember that one discovery token can represent only one etcd cluster. As long as discovery protocol on this token starts, even if it fails halfway, it must not be used to bootstrap another etcd cluster.
 
-The rest of this article will walk through the discovery process with examples that correspond to a self-hosted discovery cluster. The public discovery service, discovery.etcd.io, functions the same way, but with a layer of polish to abstract away ugly URLs, generate UUIDs automatically, and provide some protections against excessive requests. At its core, the public discovery service still uses an etcd cluster as the data store as described in this document.
+The rest of this article will walk through the discovery process with examples that correspond to a self-hosted discovery cluster.
+
+Note that this document is only for v3 discovery. Check previous document for more details on [v2 discovery][v2-discovery].
 
 ## Protocol workflow
 
 The idea of discovery protocol is to use an internal etcd cluster to coordinate bootstrap of a new cluster. First, all new members interact with discovery service and help to generate the expected member list. Then each new member bootstraps its server using this list, which performs the same functionality as -initial-cluster flag.
 
-In the following example workflow, we will list each step of protocol in curl format for ease of understanding.
+In the following example workflow, we will list each step of protocol using `etcdctl` command for ease of understanding, and we assume that `http://example.com:2379` hosts an etcd cluster for discovery service.
 
-By convention the etcd discovery protocol uses the key prefix `_etcd/registry`. If `http://example.com` hosts an etcd cluster for discovery service, a full URL to discovery keyspace will be `http://example.com/v2/keys/_etcd/registry`. We will use this as the URL prefix in the example.
+By convention the etcd discovery protocol uses the key prefix `/_etcd/registry`.
+
 
 ### Creating a new discovery token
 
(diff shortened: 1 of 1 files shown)


content/en/docs/v3.6/dev-internal/discovery_protocol.md
```diff
@@ -33,85 +36,72 @@ UUID=$(uuidgen)
...
 ```
 
 Usually the cluster size is 3, 5 or 7. Check [optimal cluster size][cluster-size] for more details.
 
 ### Bringing up etcd processes
 
-Given the discovery URL, use it as `-discovery` flag and bring up etcd processes. Every etcd process will follow this next few steps internally if given a `-discovery` flag.
-
-### Registering itself
+Set the discovery token `${UUID}` to `--discovery-token` flag, and set the endpoints of the etcd cluster backing the discovery service to `--discovery-endpoints` flag,  then v3 discovery is enabled to bootstrap the etcd cluster.
```
spzala, line 46: May be rewrite as this, for better readability? - "Set the discovery token `${UUID}` to `--discovery-token` flag, and set the endpoints of the etcd cluster backing the discovery service to `--discovery-endpoints` flag. This will enable v3 discovery to bootstrap the etcd cluster."

content/en/docs/v3.6/dev-internal/discovery_protocol.md
```diff
@@ -33,85 +36,72 @@ UUID=$(uuidgen)
...
 
-The first thing for etcd process is to register itself into the discovery URL as a member. This is done by creating member ID as a key in the discovery URL.
+Every etcd process will follow the next few steps internally if given `--discovery-token` and `--discovery-endpoints` flags.
```
spzala, line 48: Suggest to rewrite for better readability as "Every etcd process will follow the next few steps internally if `--discovery-token` and `--discovery-endpoints` flags are given."

content/en/docs/v3.6/dev-internal/discovery_protocol.md
```diff
@@ -33,85 +36,72 @@ UUID=$(uuidgen)
...
 
+If the discovery service enables client cert authentication, configure the following flags appropriately as well. They follow exactly the same usage as `etcdctl` communicating with an etcd cluster.
```
spzala, line 50: Rewrite as this? - "If the discovery service enables client cert authentication, configure the following flags. They follow exactly the same usage as using the `etcdctl` to communicate with an etcd cluster."

content/en/docs/v3.6/dev-internal/discovery_protocol.md
```diff
@@ -33,85 +36,72 @@ UUID=$(uuidgen)
...
+--discovery-cert
+--discovery-key
+--discovery-cacert
 ```
 
-### Checking the status
-
-It checks the expected cluster size and registration status in discovery URL, and decides what the next action is.
-
+If the discovery service enables role based authentication, configure the following flags appropriately as well. They follow exactly the same usage as `etcdctl` communicating with an etcd cluster.
```
spzala, line 59: Same as above.

content/en/docs/v3.6/dev-internal/discovery_protocol.md
```diff
@@ -33,85 +36,72 @@ UUID=$(uuidgen)
...
-
-If the number of registered members is bigger than the expected size N, it treats the first N registered members as the member list for the cluster. If the member itself is in the member list, the discovery procedure succeeds and it fetches all peers through the member list. If it is not in the member list, the discovery procedure finishes with the failure that the cluster has been full.
-
-In etcd implementation, the member may check the cluster status even before registering itself. So it could fail quickly if the cluster has been full.
-
-### Waiting for all members
-
-The wait process is described in detail in the [etcd API documentation][api].
-
+The default time or timeout values can also be changed using the following flags, which also follow exactly the same usage as `etcdctl` communicating with an etcd cluster.
```
spzala, line 65: Same as above for the second half of the sentence.

content/en/docs/v3.6/dev-internal/discovery_protocol.md
```diff
@@ -33,85 +36,72 @@ UUID=$(uuidgen)
...
+--discovery-keepalive-timeout
 ```
 
-It keeps waiting until finding all members.
-
-## Public discovery service
+### Registering itself
 
-CoreOS Inc. hosts a public discovery service at https://discovery.etcd.io/ , which provides some nice features for ease of use.
+The first thing what each etcd process does is to register itself into the given new cluster as a member. This is done by creating member ID as a key in the full registry key.
```
spzala, line 75: Replace "what" with "that"

content/en/docs/v3.6/dev-internal/discovery_protocol.md
```diff
@@ -33,85 +36,72 @@ UUID=$(uuidgen)
...
-	200 OK
-	400 Bad Request
-200 Body:
-	generated discovery url
+etcdctl --endpoints=http://example.com:2379 get /_etcd/registry/${UUID}/_config/size
+etcdctl --endpoints=http://example.com:2379 get /_etcd/registry/${UUID}/members
 ```
 
-The generation process in the service follows the steps from [Creating a New Discovery Token][new-discovery-token] to [Specifying the Expected Cluster Size][expected-cluster-size].
+If registered members are still not enough, it will wait for left members to appear.
```
spzala, line 90: I think we should say "...., it will wait for other members to appear" (i.e. replace left with other)?

content/en/docs/v3.6/dev-internal/discovery_protocol.md
```diff
@@ -33,85 +36,72 @@ UUID=$(uuidgen)
...
 
-### Check discovery status
+If the number of registered members is bigger than the expected size N, it treats the first N registered members as the member list for the cluster. If the member itself is in the member list, the discovery procedure succeeds, and it fetches all peers through the member list. If it is not in the member list, the discovery procedure finishes with the failure that the cluster has been full.
 
-```
-GET /${UUID}
-```
+In etcd implementation, the member may check the cluster status even before registering itself. So it could fail quickly if the cluster has been full.
```
spzala, line 94: We may just want to remove "In etcd implementation," and start the sentence as, "The member may check the..."

spzala (commented):
@ahrtr thanks for updating the doc. It looks good but few small comments.

content/en/docs/v3.6/dev-internal/discovery_protocol.md
ahrtr: Thanks & updated.

content/en/docs/v3.6/dev-internal/discovery_protocol.md
ahrtr: Thanks & updated.

content/en/docs/v3.6/dev-internal/discovery_protocol.md
ahrtr: Thanks & updated.

content/en/docs/v3.6/dev-internal/discovery_protocol.md
ahrtr: Thanks & updated.

content/en/docs/v3.6/dev-internal/discovery_protocol.md
ahrtr: Thanks & updated.

content/en/docs/v3.6/dev-internal/discovery_protocol.md
ahrtr: Thanks & updated.

content/en/docs/v3.6/dev-internal/discovery_protocol.md
ahrtr: Thanks & updated.

content/en/docs/v3.6/dev-internal/discovery_protocol.md
ahrtr: Thanks & updated.

ahrtr:
@spzala Thanks for you comments, which make sense to me!

spzala:
> @spzala Thanks for you comments, which make sense to me! All resolved.

Thank you @ahrtr

spzala (approved):
lgtm Thanks @ahrtr

content/en/docs/v3.6/dev-internal/discovery_protocol.md
```diff
@@ -33,85 +36,72 @@ UUID=$(uuidgen)
 The discovery token expects a cluster size that must be specified. The size is used by the discovery service to know when it has found all members that will initially form the cluster.
 
 ```
-curl -X PUT http://example.com/v2/keys/_etcd/registry/${UUID}/_config/size -d value=${cluster_size}
+etcdctl --endpoints=http://example.com:2379 put /_etcd/registry/${UUID}/_config/size ${cluster_size}
```
serathius, line 39: Not sure we can switch the curl to etcdctl as it is not obvious how developer would reimplement it. V2 discovery protocol was based on V2 api which was simple REST api that could be called via curl, showcasing how to implement it. Problem is that V3 api is using grpc which is binary protocol, so it's no longer obvious how we should explain it.

Could we maybe provide link to grpc proto definition and list which methods need to be implemented. It would be also great if we could provide examples by using client similar to curl however one native to grpc.

ahrtr: The developers just need to follow the guide if they want to bootstrap a new cluster using v3 discovery. Probably we should add a simple example (as a new section) to help developers to understand it.

With regard to migrating from v2 discovery to v3 discovery, previously @ptabor has a [comment](https://github.com/etcd-io/etcd/issues/13624#issuecomment-1020966905) .  Both v2 & v3 discovery are only useful at the very first bootstrapping. Once etcd members have local data, they do not depend on the discovery service any more. So it doesn't make much sense to migrate the previous v2 discovery records? 

We don't know the real data on how many users deploy the v2 discovery in their production environment. Most likely the use cases are very few. So users can manually perform the migration if they want.

What do you think?

serathius: > The developers just need to follow the guide if they want to bootstrap a new cluster using v3 discovery. Probably we should add a simple example (as a new section) to help developers to understand it.

My understanding of this documentation is not for developer to learn it, but to be able to implement their service serving discover without deploying full etcd. For example https://github.com/etcd-io/discoveryserver. I would expect that this should be cough when we were migrating tests from V2 to V3, but it didn't.
 
> With regard to migrating from v2 discovery to v3 discovery, previously @ptabor has a [comment](https://github.com/etcd-io/etcd/issues/13624#issuecomment-1020966905) . Both v2 & v3 discovery are only useful at the very first bootstrapping. Once etcd members have local data, they do not depend on the discovery service any more. So it doesn't make much sense to migrate the previous v2 discovery records?

I think so.

> We don't know the real data on how many users deploy the v2 discovery in their production environment. Most likely the use cases are very few. So users can manually perform the migration if they want.
 
I think we could look into metrics of https://github.com/etcd-io/discoveryserver which is the public etcd service discovery maintained by etcd project.

> What do you think?

Feels like we should also consider implementing V3 discovery in https://github.com/etcd-io/discoveryserver. However I would be worried to change and deploy a project that haven't seen much updates for last couple of years. 

TODO:
* Look at usage metrics of discoveryserver
* Decide if we want to implement V3 to discoveryserver
* Add tests to etcd that use discoveryserver, to make sure we don't break things in future

ahrtr: > My understanding of this documentation is not for developer to learn it, but to be able to implement their service serving discover without deploying full etcd. For example https://github.com/etcd-io/discoveryserver. I would expect that this should be cough when we were migrating tests from V2 to V3, but it didn't.

Not really per my understanding. The [discovery.etcd.io](https://github.com/coreos/discovery.etcd.io) or [discoveryserver](https://github.com/etcd-io/discoveryserver) is just a router, which routes the client requests to backend etcd cluster. Please note that users do not necessarily to use the public discovery API, they can just deploy an etcd cluster as the discovery service themselves. 



> I think we could look into metrics of https://github.com/etcd-io/discoveryserver which is the public etcd service discovery maintained by etcd project.

This is a good point. But just as I mentioned above, users do not necessarily use the public discovery API. But anyway, we can take a look at the metrics firstly. Do you or @ptabor know the info on the backend etcd server supporting the public discovery service? 



> * Decide if we want to implement V3 to discoveryserver

I don't think we should implement a public discovery service for V3 discovery. But it's open to discussion.

serathius: I read some documentation, you are right it's just a router to etcd server underneath. Question should we also implement it to support v3 discovery to avoid removing this feature?

ahrtr: A couple of comments:
1. There are already several [e2e cases](https://github.com/etcd-io/etcd/blob/main/tests/e2e/discovery_test.go) to cover the v2 discovery. We can remove them in 3.7 or 3.8 when clientv2 is completely removed.
2. With regard to the question whether should we support public v3 discovery service, we can discuss it in separate session after we investigate the existing data in pubic v2 discovery service. Personally I don't think it's a priority for now. 
3. I will try to add a new example section into the doc in a separate PR.  

This PR should be good.

ahrtr:
Can we get this PR merged so that I can go ahead to submit a separate PR to add an example on how to use v3 discovery?

ahrtr:
@spzala @serathius @ptabor Can we get this PR merged? thx

spzala:
> @spzala @serathius @ptabor Can we get this PR merged? thx

@ahrtr yes, your approach sounds good to me. Thanks!

ahrtr:
> > @spzala @serathius @ptabor Can we get this PR merged? thx
> 
> @ahrtr yes, your approach sounds good to me. Thanks!

Thanks @spzala .  I have no permission on this repo (`website`), please help to merge this PR if there is no any objection. Thanks. cc @serathius 

I will enhance the doc in separate PRs. Note that we also need to update https://etcd.io/docs/v3.5/op-guide/clustering/#discovery.

spzala:
> > > @spzala @serathius @ptabor Can we get this PR merged? thx
> > 
> > 
> > @ahrtr yes, your approach sounds good to me. Thanks!
> 
> Thanks @spzala . I have no permission on this repo (`website`), please help to merge this PR if there is no any objection. Thanks. cc @serathius
> 
> I will enhance the doc in separate PRs. Note that we also need to update https://etcd.io/docs/v3.5/op-guide/clustering/#discovery.

@ahrtr sounds good. Will be working on your permission as well. Thanks!

PR #586, opened 2022-05-20
etcdserver: alter the documentation for the quota-backend-bytes variable

### LINKED ISSUE #585: Alter the documentation for the quota-backend-bytes variable
Hello,
Sorry to be nitpicky; I think there's room for misunderstanding in the current documentation.

The current [documentation](https://etcd.io/docs/v3.6/dev-guide/limit/#storage-size-limit) declares that the default value for the `--quota-backend-bytes` is 2GB, but when I look at the [default value](https://github.com/etcd-io/etcd/blob/main/server/storage/quota.go#L31-L34=), it's equal to 2GiB. Usually, this should not be an issue, but it might be best to be verbose since this value is in bytes.

Another issue with this variable is in the [configuration flags page](https://etcd.io/docs/v3.6/op-guide/configuration/). It would be better to declare what "low space quota" means.

`(0 defaults to low space quota).`

I will create a PR with my suggested changes. :) 
Thanks for the project!

necatican:
To be more verbose and clear out possible misunderstandings. It is best to know the exact values for flags.

Fixes #585

I haven't checked if we have this issue in other places. However, the source code also refers to the value as 2GB. 

I can edit my commit to add these lines to previous versions.

### DIFF
--- a/content/en/docs/v3.5/dev-guide/limit.md
+++ b/content/en/docs/v3.5/dev-guide/limit.md
@@ -10,4 +10,4 @@ etcd is designed to handle small key value pairs typical for metadata. Larger re
 
 ## Storage size limit
 
-The default storage size limit is 2GB, configurable with `--quota-backend-bytes` flag. 8GB is a suggested maximum size for normal environments and etcd warns at startup if the configured value exceeds it.
+The default storage size limit is 2GiB, configurable with `--quota-backend-bytes` flag. 8GiB is a suggested maximum size for normal environments and etcd warns at startup if the configured value exceeds it.
--- a/content/en/docs/v3.5/op-guide/configuration.md
+++ b/content/en/docs/v3.5/op-guide/configuration.md
@@ -57,7 +57,7 @@ Flags are presented below using the format `--flag-name DEFAULT_VALUE`.
 --max-wals '5'
   Maximum number of wal files to retain (0 is unlimited).
 --quota-backend-bytes '0'
-  Raise alarms when backend size exceeds the given quota (0 defaults to low space quota).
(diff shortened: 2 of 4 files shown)


content/en/docs/v3.5/dev-guide/limit.md
```diff
@@ -10,4 +10,4 @@ etcd is designed to handle small key value pairs typical for metadata. Larger re
 
 ## Storage size limit
 
-The default storage size limit is 2GB, configurable with `--quota-backend-bytes` flag. 8GB is a suggested maximum size for normal environments and etcd warns at startup if the configured value exceeds it.
+The default storage size limit is 2GiB, configurable with `--quota-backend-bytes` flag. 8GiB is a suggested maximum size for normal environments and etcd warns at startup if the configured value exceeds it.
```
jmhbnz, line 13: This looks like a good clarification to add, thanks for raising.  I would suggest we update all versions of the documentation with this limit present at the same time for consistency.

jmhbnz (approved):
LGTM

@ahrtr , @spzala - It's been quite a while since this was opened so I say we merge as is without addressing my comment above, I am happy to raise a quick follow-up to fix the other versions.

ahrtr (approved):
LGTM

Probably we should update 3.4 doc as well.

jmhbnz:
Hey @spzala - Can you please consider merging this as is - I will follow up with fix for other versions, thanks!

spzala:
@jmhbnz @necatican can you please fix merge conflict. Thanks!

jmhbnz:
Closing as completed via #665 and #667

PR #590, opened 2022-06-14
Fix typos in upgrade guides.

cristiangreco:
deprcates -> deprecates


spzala (approved):
lgtm. thanks for the clean sweep @cristiangreco

Issue #595, opened 2022-07-12
Set up Google Analytics 4 (GA4)


chalin:
- [x] Create a GA4 tag (measurement ID) --> G-VF9H3EHMDK. For details see comments below
  - **Note**: the UA and GA4 tags are now [connected]
- [x] Add ~or upgrade to~ use of `gtag.js` for configuring the new GA4 site tag - #597
- [x] Confirm that the GA4 site tag is receiving events and posting data

To learn more, including links to other resources, see:

- https://github.com/cncf/techdocs/issues/108

/cc @nate-double-u 

[connected]: https://support.google.com/analytics/answer/9973999

chalin:
GA4 tag is receiving data -- this is a few minutes after the deploy of #597:

> <img width="273" alt="image" src="https://user-images.githubusercontent.com/4140793/178996284-8d141dac-d15e-49e0-876b-aef5f46dd82a.png">

PR #597, opened 2022-07-13
Analytics: enable GA4 in addition to UA

chalin:
- Contributes to #595 by configuring the GA4 site tag _in addition_ to the UA tag (which is currently being configured via the Hugo internal template).

  Having both is an acceptable way to transition to GA4: for details see **Option 1** of https://developers.google.com/analytics/devguides/migration/measurement/add-ga4?technology=analyticsjs#set_up_the_basic_page_tag
- Notes:
  -  We won't be able to see the effect of this change until it is deployed to the production server.
  - We could have _moved_ the UA ID to the end of the `params.googleAnalytics` parameter, but that would have disabled the [page feedback] feature, which is not what we want.

[page feedback]: https://www.docsy.dev/docs/adding-content/feedback/#user-feedback


nate-double-u (approved):
/lgtm
thanks @chalin!

spzala (approved):
Thanks @chalin @nate-double-u

PR #600, opened 2022-07-19
Add alert rules

spzala:
The added file  in this PR was missing in the v3.5 doc. It was there
in the previous releases and used by the Kube Prometheus Stack.

Related #556

Signed-off-by: Sahdev Zala spzala@us.ibm.com

### DIFF
--- a/content/en/docs/v3.5/op-guide/etcd3_alert.rules.yml
+++ b/content/en/docs/v3.5/op-guide/etcd3_alert.rules.yml
@@ -1,0 +1,148 @@
+# these rules synced manually from https://github.com/etcd-io/etcd/blob/release-3.5/contrib/mixin/mixin.libsonnet
+groups:
+- name: etcd
+  rules:
+  - alert: etcdMembersDown
+    annotations:
+      description: 'etcd cluster "{{ $labels.job }}": members are down ({{ $value }}).'
+      summary: etcd cluster members are down.
+    expr: |
+      max without (endpoint) (
+        sum without (instance) (up{job=~".*etcd.*"} == bool 0)
+      or
+        count without (To) (
+          sum without (instance) (rate(etcd_network_peer_sent_failures_total{job=~".*etcd.*"}[120s])) > 0.01
+        )
+      )
+      > 0
+    for: 10m
+    labels:
+      severity: critical
+  - alert: etcdInsufficientMembers
+    annotations:
+      description: 'etcd cluster "{{ $labels.job }}": insufficient members ({{ $value }}).'
+      summary: etcd cluster has insufficient number of members.
+    expr: |
+      sum(up{job=~".*etcd.*"} == bool 1) without (instance) < ((count(up{job=~".*etcd.*"}) without (instance) + 1) / 2)
+    for: 3m
+    labels:
+      severity: critical
+  - alert: etcdNoLeader
+    annotations:
+      description: 'etcd cluster "{{ $labels.job }}": member {{ $labels.instance }} has no leader.'
+      summary: etcd cluster has no leader.
+    expr: |
+      etcd_server_has_leader{job=~".*etcd.*"} == 0
+    for: 1m
+    labels:
+      severity: critical
+  - alert: etcdHighNumberOfLeaderChanges
+    annotations:
+      description: 'etcd cluster "{{ $labels.job }}": {{ $value }} leader changes within the last 15 minutes. Frequent elections may be a sign of insufficient resources, high network latency, or disruptions by other components and should be investigated.'
+      summary: etcd cluster has high number of leader changes.
+    expr: |
+      increase((max without (instance) (etcd_server_leader_changes_seen_total{job=~".*etcd.*"}) or 0*absent(etcd_server_leader_changes_seen_total{job=~".*etcd.*"}))[15m:1m]) >= 4
+    for: 5m
+    labels:
+      severity: warning
+  - alert: etcdHighNumberOfFailedGRPCRequests
+    annotations:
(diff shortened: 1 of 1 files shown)


spzala:
Fyi, As it's mentioned in the file as a comment, I have generated this file from https://github.com/etcd-io/etcd/blob/release-3.5/contrib/mixin/mixin.libsonnet Thanks!

ahrtr:
Thanks @spzala .

The concern is that we need to make sure [etcd3_alert.rules.yml](https://github.com/etcd-io/website/pull/600/files#diff-82858d605ec64219a8873fd3548e440cfb5ee7fc2a8a94cdf6c607ee994129ea) is in sync with [mixin.libsonnet](https://github.com/etcd-io/etcd/blob/main/contrib/mixin/mixin.libsonnet). 

Probably it would be better to only maintain [mixin.libsonnet](https://github.com/etcd-io/etcd/blob/main/contrib/mixin/mixin.libsonnet)?

ahrtr:
cc  @wking @simonpasquier @Elbehery @serathius @ptabor @hexfusion

Elbehery:
> etcd3_alert.rules.yml

+1 we should only maintain [mixin.libsonnet](https://github.com/etcd-io/etcd/blob/main/contrib/mixin/mixin.libsonnet) as the single source of truth :)

content/en/docs/v3.5/op-guide/etcd3_alert.rules.yml
```diff
@@ -0,0 +1,148 @@
...
+      > 0.25
+    for: 10m
+    labels:
+      severity: warning
+  - alert: etcdBackendQuotaLowSpace
+    annotations:
+      description: 'etcd cluster "{{ $labels.job }}": database size exceeds the defined quota on etcd instance {{ $labels.instance }}, please defrag or increase the quota as the writes to etcd will be disabled when it is full.'
+      summary: etcd cluster database is running full.
+    expr: |
+      (etcd_mvcc_db_total_size_in_bytes/etcd_server_quota_backend_bytes)*100 > 95
```
Elbehery, line 136: this has been changed at https://github.com/etcd-io/etcd/pull/14237

Elbehery:
> Fyi, As it's mentioned in the file as a comment, I have generated this file from https://github.com/etcd-io/etcd/blob/release-3.5/contrib/mixin/mixin.libsonnet Thanks!

so this is the result of `make manifests`, correct ?

content/en/docs/v3.5/op-guide/etcd3_alert.rules.yml
Elbehery:
```suggestion
(last_over_time(etcd_mvcc_db_total_size_in_bytes[5m]) / last_over_time(etcd_server_quota_backend_bytes[5m]))*100 > 95
```

serathius:
I'm also for single source of truth. Preferably website should just redirect users to [mixin.libsonnet](https://github.com/etcd-io/etcd/blob/main/contrib/mixin/mixin.libsonnet).

spzala:
@ahrtr @serathius @Elbehery thanks for the quick review comments. So, to add more details and address comments: 
1) Yes, I also mentioned in the related issue, https://github.com/etcd-io/website/pull/556, that we should be using [mixin.libsonnet](https://github.com/etcd-io/etcd/blob/main/contrib/mixin/mixin.libsonnet). As mentioned there, the Kube Prometheus Stack seems still using the etcd3_alert.rules.yml file Also, we are not sure if anyone else using it. 
2) The file was generated from the `mixin.libsonnet` from the v3.5 branch as mentioned in the file. (using `make` and Kubernetes-mixin [instructiion](https://github.com/kubernetes-monitoring/kubernetes-mixin)) Since this is added to v3.5 doc, I think we won't be doing any maintenance here except one time adding this file. 
3) We can tell Kube Prometheus Stack that from v3.6 we won't be supporting etcd3_alert.rules.yml  any more. We can also tell them same for v3.5 (and don't add this file) but I thought adding this file is harmless and my guess is no maintenance as I mentioned above. I am good to go with what you all suggest. Thanks!

content/en/docs/v3.5/op-guide/etcd3_alert.rules.yml
spzala: As it says in the file, it was generated from the v3.5 mixin,  https://github.com/etcd-io/etcd/blob/release-3.5/contrib/mixin/mixin.libsonnet

spzala:
> Thanks @spzala .
> 
> The concern is that we need to make sure [etcd3_alert.rules.yml](https://github.com/etcd-io/website/pull/600/files#diff-82858d605ec64219a8873fd3548e440cfb5ee7fc2a8a94cdf6c607ee994129ea) is in sync with [mixin.libsonnet](https://github.com/etcd-io/etcd/blob/main/contrib/mixin/mixin.libsonnet).
> 
> Probably it would be better to only maintain [mixin.libsonnet](https://github.com/etcd-io/etcd/blob/main/contrib/mixin/mixin.libsonnet)?

@ahrtr I mentioned above but to be more clear, the file was generated from the v3.5 branch, [last_over_time ](https://github.com/etcd-io/etcd/blob/main/contrib/mixin/mixin.libsonnet) so we don't need to sync and keep it updated with the master branch mixin.

ahrtr:
Thanks @spzala for the clarification.  It makes sense to me, because as you mentioned it's just one time operation for 3.5 and from v3.6 we won't be supporting etcd3_alert.rules.yml any more.

ahrtr (approved):
LGTM

Elbehery:
Thanks for clarifying this @spzala 

LGTM

spzala:
Thanks @Elbehery

PR #601, opened 2022-08-03
Update versioning support

spzala:
This PR updates versioning support per the discussion under etcd
issue - https://github.com/etcd-io/etcd/issues/13912

Signed-off-by: Sahdev Zala <spzala@us.ibm.com>

### DIFF
--- a/content/en/docs/v3.3/op-guide/versioning.md
+++ b/content/en/docs/v3.3/op-guide/versioning.md
@@ -11,9 +11,9 @@ This document describes the versions supported by the etcd project.
 etcd versions are expressed as **x.y.z**, where **x** is the major version, **y** is the minor version, and **z** is the patch version, following [Semantic Versioning](https://semver.org/) terminology.
 New minor versions may add additional features to the API.
 
-The etcd project maintains release branches for the current version and two minor releases, 3.5, 3.4 and 3.3 currently.
+The etcd project maintains release branches for the current version and previous release. For example, when v3.5 is the current version, v3.4 is supported. When v3.6 is released, v3.4 goes out of support.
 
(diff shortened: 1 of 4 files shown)


spzala:
cc @ahrtr @serathius @ptabor

ahrtr:
Thanks @spzala for clearly clarifying this in our official doc.

Usually a project needs to support three latest versions, so does K8s and etcd.  But etcd doesn't keep the release promise just as the issue https://github.com/etcd-io/etcd/issues/13912 pointed out. Personally I agree the proposal of only supporting two versions (such as 3.5 and 3.4 for now), reasons & points:
1. With limited active maintainers, we have no resource to support three versions for now;
2. All active Kubernetes versions are using etcd 3.5. Of course, the public managed K8s platforms (such as EKS, GKE, etc.) are still running K8s 1.21/etcd 3.4. But no any K8s platform is using etcd 3.3, so It's safe to give up etcd 3.3. 
3. Before we release etcd 3.6, we really need to stabilize etcd 3.5 so that all of us (including all contributors and users) have enough confidence to completely migrating to 3.5. So there is no rush to release 3.6.
4. If anybody doesn't agree with the proposal, then we can revisit the proposal when we release etcd 3.7.  It means the proposal should be good enough before we release etcd 3.7. The reason is that both etcd 3.3 and 3.4 are out of support for a long time, of course 3.4 is much better after we release 3.4.19 and 3.4.20 (coming soon); so we really want all use cases converge to 3.5 as soon as possible.

ahrtr (approved):
LGTM

spzala:
@ahrtr thanks for the added context. It will be helpful for anyone referencing this PR. And yes, absolutely, we can revisit the proposal and current policy as per the future need and when we have increased bandwidth in the team.

PR #602, opened 2022-08-08
Update recovery.md

goncalocribeiro:
Data dir must be set when restoring and when starting an etcd node


jmhbnz:
Closing this as a duplicate of #603

PR #611, opened 2022-09-12
Add stability level for etcd metrics and promote some important etcd metrics as stable

CatherineF-dev:
Fixes https://github.com/etcd-io/etcd/issues/14448

### DIFF
--- a/content/en/docs/v3.6/metrics.md
+++ b/content/en/docs/v3.6/metrics.md
@@ -24,14 +24,14 @@ These metrics describe the status of the etcd server. In order to detect outages
 
 All these metrics are prefixed with `etcd_server_`
 
-| Name                      | Description                                              | Type    |
-|---------------------------|----------------------------------------------------------|---------|
-| has_leader                | Whether or not a leader exists. 1 is existence, 0 is not.| Gauge   |
-| leader_changes_seen_total | The number of leader changes seen.                       | Counter |
-| proposals_committed_total | The total number of consensus proposals committed.       | Gauge   |
-| proposals_applied_total   | The total number of consensus proposals applied.         | Gauge   |
-| proposals_pending         | The current number of pending proposals.                 | Gauge   |
-| proposals_failed_total    | The total number of failed proposals seen.               | Counter |
+| Name                      | Description                                              | Type    | Stability Level |
+|---------------------------|----------------------------------------------------------|---------|-----------------|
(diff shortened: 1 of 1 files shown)


CatherineF-dev:
Not sure metrics peer* are stable or not.

CatherineF-dev:
cc @ahrtr

ahrtr:
Thanks for the PR.

It's clearly stated that `The metrics under the etcd prefix are for monitoring and alerting. They are stable high level metrics.`, please see [#etcd-namespace-metrics](https://etcd.io/docs/v3.5/metrics/#etcd-namespace-metrics); and `The metrics under the etcd_debugging prefix are for debugging. They are very implementation dependent and volatile. They might be changed or removed without any warning in new etcd releases.`, please see [#etcd_debugging-namespace-metrics](https://etcd.io/docs/v3.5/metrics/#etcd_debugging-namespace-metrics).

Probably we don't need to add stable/experimental for each single metrics. cc @spzala @serathius @ptabor for option.

Why do you mark some metrics prefixed with `etdd_network` as experimental?

CatherineF-dev:
Will update network metrics as stable.  

So, is it okay to re-state metric stability level? 
If we add an additional column stability, it's easier to understand. Also, this stability level is consistent with OSS kubernetes and kube-state-metrics stability level.

spzala (commented):
Thanks @CatherineF-dev and +1 @ahrtr I am not opposed to the changes but considering the way etcd Metrics doc is structured, I think adding a new column is redundant and clutter the doc as any debugging/experimental metrics are always listed in a debugging section. The Metrics doc also clearly says it so I would favor to close the PR unless we want to change the doc structure which probably is not needed. In Kubernetes, the doc structure is different and all the metrics are listed in single table(s) and so stable and experimental separation is needed. Thanks!

CatherineF-dev:
Okay, how about highlighting `stable high` in  [`They are stable high level metrics.`](https://etcd.io/docs/v3.5/metrics/#etcd-namespace-metrics)

So that etcd contributors are less likely to change metric name or delete metric labels.

spzala:
I think it's not needed but it's fine with me to emphasize stable if you find it helpful and if other reviewers are okay with it. Thanks!

jmhbnz:
I'm going to close this one as it has been sitting open for almost six months and two maintainers have taken a view that this change is not required. I tend to agree given we already state the following for all our `etcd` namespace metrics:

> _They are stable high level metrics._  

We can revisit this in future if any confusion emerges about what etcd metrics are stable.

PR #613, opened 2022-09-24
doc: typo fix

robel-yemane:
typo in the binary name etcdctl

Signed-off-by: Robel Yemane <ryhgb03@gmail.com>


spzala (commented):
Thanks but etcdutl is not a typo, it's an utility. As you can see it's used multiple times in the doc. Also, you can read more here - https://github.com/etcd-io/etcd/tree/main/etcdctl#snapshot-restore-options-filename

robel-yemane:
🤦 , it is indeed. Cheers @spzala

PR #615, opened 2022-09-26
Update etcd txn API doc to clarify the inner revision in TxnResponse

ahrtr:
Clarify https://github.com/etcd-io/etcd/issues/14490 

Signed-off-by: Benjamin Wang <wachao@vmware.com>

cc @spzala @serathius @ptabor

### DIFF
--- a/content/en/docs/v3.6/learning/api.md
+++ b/content/en/docs/v3.6/learning/api.md
@@ -323,6 +323,10 @@ message ResponseOp {
 }
 ```
 
+The `ResponseHeader` included in each inner response shouldn't be interpreted in any way.
+If clients need to get the latest revision, then they should always check the top level `ResponseHeader` in `TxnResponse`.
+
+
 ## Watch API
 
 The `Watch` API provides an event-based interface for asynchronously monitoring changes to keys. An etcd3 watch waits for changes to keys by continuously watching from a given revision, either current or historical, and streams key updates back to the client.
--- a/content/en/docs/v3.4/learning/api.md
+++ b/content/en/docs/v3.4/learning/api.md
@@ -323,6 +323,10 @@ message ResponseOp {
 }
 ```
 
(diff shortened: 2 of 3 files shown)


content/en/docs/v3.6/learning/api.md
```diff
@@ -323,6 +323,10 @@ message ResponseOp {
 }
 ```
 
+For `response_range` and `response_delete_range` responses in `TxnResponse`, the `revision` included in the inner response header is just the revision when generating the response, so it might be one less than the `revision` included
```
socketpair, line 326: Actually all responses, without clarification. And also without clarification about "one less".

I would write something like:

```
Response headers returned in TXN should not be interpreted in any way, and should be completely ignored.
```

ahrtr: Thanks for the feedback. Just updated the doc. 

It isn't decided to deprecate the inner `ResponseHeader` for now. Please note that each response has exactly the same format as non-TXN response. 

The `revision` included in each response might be still useful, such as checking whether the deletion is successful or not when there is no PUT request before it.

ahrtr: Makes sense to me & updated & thx.

Note that there is no plan to remove the `ResponseHeader` in each inner response. See https://github.com/etcd-io/etcd/issues/14490#issuecomment-1257086651

ahrtr:
@spzala @serathius Please take a look.

vm86:
I'm writing on behalf of @socketpair :) 
LGTM!

socketpair:
Agreed

spzala (approved):
Good clarification. 
lgtm Thanks @ahrtr @socketpair

ahrtr:
@spzala @serathius Please help to merge this PR if there is no any objection. thx

spzala:
Sure @ahrtr

PR #616, opened 2022-09-27
Document compacted revision hash check

serathius:
cc @ahrtr @ptabor @spzala 
When documentation is approved for v3.5 I will also copy it to v3.6 in separate PR.


spzala (commented):
@serathius nice, thanks! One comment here, we should add this file under `https://etcd.io/docs/v3.4/op-guide/` similar to what we have for v3.5 https://etcd.io/docs/v3.5/op-guide/data_corruption/ unless you have kept it to the root doc for any reason.

serathius:
Done.

spzala (approved):
awesome, thanks @serathius

PR #627, opened 2022-11-10
update doc v3.5 grpc_naming.md

cleverhu:
The correct format  is `clientv3 "go.etcd.io/etcd/client/v3"` instead of `"go.etcd.io/etcd/clientv3"`
image


spzala (approved):
lgtm
Thanks @cleverhu

PR #633, opened 2022-12-02
grpc_naming.md round-robin strategy

ramil600:
when using new resolver from, default grpc load balancing strategy: pick first will be used. It would be nice to point out how to use round-robin load balancing.

Signed-off-by: Ramil Mirhasanov <ramil600@yahoo.com>

### DIFF
--- a/content/en/docs/v3.6/dev-guide/grpc_naming.md
+++ b/content/en/docs/v3.6/dev-guide/grpc_naming.md
@@ -44,6 +44,14 @@ The etcd client's `endpoints.Manager` method can also register new endpoints wit
 em := endpoints.NewManager(client, "foo/bar/my-service")
 err := em.AddEndpoint(context.TODO(),"foo/bar/my-service/e1", endpoints.Endpoint{Addr:"1.2.3.4"});
 ```
+To enable round-robin load balancing when dialing service with multiple endpoints, you can set up you connection with grpc
+ internal round-robin load balancer:
+
+ ```go
+
+ 	conn, gerr := grpc.Dial("etcd:///foo", grpc.WithResolvers(etcdResolver),
+ 		grpc.WithDefaultServiceConfig(`{"loadBalancingPolicy":"round_robin"}`))
+ ```
 
 ### Deleting an endpoint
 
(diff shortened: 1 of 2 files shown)


ptabor:
LGTM. But ideally we should have a test if we are making it an officially supported use-case. 
I don't recall test using the Round Robin policy.

jmhbnz:
> LGTM. But ideally we should have a test if we are making it an officially supported use-case. I don't recall test using the Round Robin policy.

Good point!  I've opened a draft here for a new case https://github.com/etcd-io/etcd/pull/15577.   We can finish and merge that if there is broader consensus that it's worthwhile to support round robin.

content/en/docs/v3.6/dev-guide/grpc_naming.md
```diff
@@ -44,6 +44,14 @@ The etcd client's `endpoints.Manager` method can also register new endpoints wit
 em := endpoints.NewManager(client, "foo/bar/my-service")
 err := em.AddEndpoint(context.TODO(),"foo/bar/my-service/e1", endpoints.Endpoint{Addr:"1.2.3.4"});
 ```
+To enable round-robin load balancing when dialing service with multiple endpoints, you can set up you connection with grpc
+ internal round-robin load balancer:
+
+ ```go
+
+ 	conn, gerr := grpc.Dial("etcd:///foo", grpc.WithResolvers(etcdResolver),
```
ahrtr, line 52: Actually Resolver is an experimental gRPC API, and we need to remove the dependency on it. FYI. https://github.com/etcd-io/etcd/issues/15145

I am OK to merge this PR firstly, and update it when https://github.com/etcd-io/etcd/issues/15145 is done. 

cc @spzala

ahrtr (approved):
LGTM with a comment, thank you @jmhbnz 

cc @spzala

content/en/docs/v3.6/dev-guide/grpc_naming.md
spzala: @ahrtr thanks and that sounds good to me too!

spzala (approved):
Thanks @ramil600 @jmhbnz

PR #642, opened 2023-01-26
Update config options

spzala:
The current way of listing flags can be incomplete and incorrect due to ongoing development changes. We should link to the etcd help for users to get the latest flags. 

Signed-off-by: Sahdev Zala <spzala@us.ibm.com>

### DIFF
--- a/content/en/docs/v3.4/op-guide/configuration.md
+++ b/content/en/docs/v3.4/op-guide/configuration.md
@@ -1,5 +1,5 @@
 ---
-title: Configuration flags
+title: Configuration options
 weight: 4050
 description: "etcd configuration: files, flags, and environment variables"
 ---
@@ -16,6 +16,8 @@ The format of environment variable for flag `--my-flag` is `ETCD_MY_FLAG`. It ap
 The [official etcd ports][iana-ports] are 2379 for client requests and 2380 for peer communication. The etcd ports can be set to accept TLS traffic, non-TLS traffic, or both TLS and non-TLS traffic.
 
 To start etcd automatically using custom settings at startup in Linux, using a [systemd][systemd-intro] unit is highly recommended.
+
+The list of flags provided below may not be up-to-date due to ongoing development changes. For the latest available flags, run `etcd --help` or refer to the [etcd help][].
 
 ## Member flags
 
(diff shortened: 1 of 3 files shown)


content/en/docs/v3.4/op-guide/configuration.md
```diff
@@ -17,6 +17,8 @@ The [official etcd ports][iana-ports] are 2379 for client requests and 2380 for
 
 To start etcd automatically using custom settings at startup in Linux, using a [systemd][systemd-intro] unit is highly recommended.
 
+The list of flags provided below is for example purposes. It may not be up-to-date. For the latest available flags, run `etcd --help` or refer to the [etcd help][].
```
ahrtr, line 20: We still need to make sure the doc is in sync with the help.go, but there maybe some delay. So I think the flags provided below are still supposed to be a full list instead of examples. WDYT? 

```suggestion
The list of flags provided below may not be up-to-date due to ongoing development changes. For the latest available flags, run `etcd --help` or refer to the [etcd help][].
```

content/en/docs/v3.4/op-guide/configuration.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Configuration flags
+title: Configuration options
```
ahrtr, line 2: Minor comment, I am not sure whether we need to change `flags` to `options`, because `flag` or `flags` is used in the entire doc.

ahrtr:
Thanks @spzala . Overall looks good to me with a couple of minor comments.

content/en/docs/v3.4/op-guide/configuration.md
spzala: @ahrtr +1 I like the tweak here. Will update. Thanks!

content/en/docs/v3.4/op-guide/configuration.md
spzala: The reason I thought we should change to `options` because there are three ways we mention for configuration: config flags, config file, and env variables. So, I found options more broader vs flags. I am okay to change it, if you still think so? Thanks!

ahrtr: Sounds good to me.

spzala: Thanks for the quick look :) @ahrtr

ahrtr (approved):
LGTM

Thanks @spzala

PR #568, opened 2022-03-15
fix missing CHANGELOG link

siansiansu:
Hi reviewers,

This minimal PR is to fix a wrong hyperlink for `CHANGELOG-3.5.md`
Please help to review it.

Original
```
https://github.com/etcd-io/etcd/blob/main/CHANGELOG-{{< psubstr version 1 >}}.md
```

In-coming change
```
https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-{{< psubstr version 1 >}}.md
```

Please click this [link](https://github.com/etcd-io/etcd/blob/main/CHANGELOG/CHANGELOG-3.5.md) to check `CHANGELOG-3.5.md`

Thanks!

PR #649, opened 2023-03-04
Fix broken v3.6 CHANGELOG link in ops guide

jmhbnz:
Quick follow-on from #568 to make sure we also fix 3.6 docs.

No other versions of docs are impacted.


ahrtr (approved):
Thanks @jmhbnz

PR #662, opened 2023-04-17
Update the RBAC page

mitake:
- --peer-cert-allowed-cn isn't related to RBAC and described in security.md, so deleting it from the page should be better
- --no-password option isn't described so I'm adding it

cc @spzala


spzala (approved):
lgtm 
Thanks @mitake

mitake:
Thanks @jmhbnz and @spzala

PR #665, opened 2023-04-20
Website doc correction for data size units

shayyxi:


jmhbnz (changes requested):
Thanks for fixing this  @shayyxi!

Just one request, can you please squash the commits to one?

Issue #573, opened 2022-03-28
Want to publish a helm chart deployment tutorial


zhangtong007:

jmhbnz:
Filling in the installation gaps on our website to cover helm would be a great first issue: https://etcd.io/docs/v3.6/install/#installation-on-kubernetes-using-a-statefulset-or-helm-chart 

We would need to decide if we will maintain a chart ourselves or just suggest people use a third party chart like https://bitnami.com/stack/etcd/helm.

Given constraints in terms of maintaining existing codebase and trying not to add new code we need to maintain it might be best to suggest a third party chart for now.

jmhbnz:
Completed via #670, thanks @chrsalx

PR #670, opened 2023-04-20
updates intallation docs for 3.5 and 3.6

chrsalx:
From #573 

This PR uptakes installation docs. The docs will now forward the reader looking to install etcd thorough Heml to Bitnami's documentation.


chrsalx:
@jmhbnz

content/en/docs/v3.5/install.md
```diff
@@ -79,7 +79,7 @@ TBD---Help Wanted
 
 ## Installation on Kubernetes, using a statefulset or helm chart
 
-TBD---Help Wanted
+You can follow the instructions provided by [Bitnami's etcd Helm chart]
```
jmhbnz, line 82: Suggest a slight change to state why we are directing readers to the Bitnami chart. Something like:

> The etcd project does not currently maintain a helm chart, however you can follow the instructions provided by [Bitnami's etcd Helm chart].

content/en/docs/v3.6/install.md
```diff
@@ -79,7 +79,7 @@ TBD---Help Wanted
 
 ## Installation on Kubernetes, using a statefulset or helm chart
 
-TBD---Help Wanted
+You can follow the instructions provided by [Bitnami's etcd Helm chart]
```
jmhbnz, line 82: As above, think we should just say why we are directing the user towards the Bitnami chart.

jmhbnz (commented):
Thanks for fixing this @chrsalx! Having something covering this installation method even a third party chart reference is better than what we currently have which is nothing at all.

Bitnami are a popular and mature chart provider so I think this is a good choice.

Just one small bit of feedback to address as shown below then I will be happy to approve this.

spzala (approved):
lgtm 
Thanks @chrsalx

jmhbnz (approved):
LGTM thanks @chrsalx

PR #676, opened 2023-04-27
Update the api guarantees based on Kubernetes-etcd implicit contract document

serathius:
Document https://docs.google.com/document/d/1NUZDiJeiIH5vo_FMaTWf0JtrQKCx0kpEaIIuPoj9P6A/edit?usp=sharing

Some concepts don't match one to one K8s only requires a subset of things guaranteed by etcd.

cc @ahrtr @ptabor @logicalhan @jpbetz @thockin @deads2k @wojtek-t


content/en/docs/v3.6/learning/api_guarantees.md
```diff
@@ -1,75 +1,60 @@
...
+description: API guarantees made by etcd
 ---
 
-etcd is a consistent and durable key value store with [mini-transaction][txn]
-support. The key value store is exposed through the KV APIs. etcd tries to
-ensure the strongest consistency and durability guarantees for a distributed
-system. This specification enumerates the KV API guarantees made by etcd.
+etcd is a consistent and durable key value store.
+The key value store is exposed through the KV and Watch [gRPC Services].
+etcd ensures the strongest consistency and durability guarantees for a distributed system.
```
serathius, line 9: This is a little on the nose, as there are stronger consistency models than strict serializability (at last in theory).

serathius:
ping @ahrtr @ptabor

content/en/docs/v3.6/learning/api_guarantees.md
serathius: For now leaving it as it is.

serathius:
Part of https://github.com/etcd-io/etcd/issues/15820

content/en/docs/v3.6/learning/api_guarantees.md
```diff
@@ -85,18 +70,36 @@ most current value. Without linearizability guarantee, the returned value,
...
+  within a compaction window. Watch may be terminated if it's no longer in
+  history window.
+* Atomic - a list of events is guaranteed to encompass complete revisions.
+  Updates in the same revision over multiple keys will not be split over several
+  lists of events.
+* Resumable - A broken watch can be resumed by establishing a new watch starting
+  after the last revision received in a watch event before the break, so long as
+  the revision is in the history window.
+* Bookmarkable - Progress notification events guarantee that all events up to a
+  revision have been already delivered.
```
ahrtr, line 97: I don't think `Resumable` and `Bookmarkable` are at the same level as other guarantees; instead, they can be inferred from other meta items. Suggest to either remove them or change it to something like below,
```suggestion

From the above guarantees, the following two items can be deduced,
* Resumable - A broken watch can be resumed by establishing a new watch starting
  after the last revision received in a watch event before the break, so long as
  the revision is in the history window.
* Bookmarkable - Progress notification events guarantee that all events up to a
  revision have been already delivered.
```

content/en/docs/v3.6/learning/api_guarantees.md
```diff
@@ -1,75 +1,60 @@
...
-revision of the operation. The revision can be used as a logical clock for key
-value store. A key value pair that has a larger revision is modified after a key
-value pair with a smaller revision. Two key value pairs that have the same
-revision are modified by an operation "concurrently".
+### Durability
+
+Any completed operations are durable. All accessible data is also durable data.
+A read will never return data that has not been made durable.
+
+### Strict serializability
```
ahrtr, line 43: ACID are four fundamental attributes of transactional database. No matter what consistency modes (e.g. linearizability, serializablity, etc.) we are describing here, they are just a type of consistency mode. Please consider to revert to original structure.

content/en/docs/v3.6/learning/api_guarantees.md
```diff
@@ -1,75 +1,60 @@
...
-description: KV API guarantees made by etcd
+description: API guarantees made by etcd
 ---
 
-etcd is a consistent and durable key value store with [mini-transaction][txn]
-support. The key value store is exposed through the KV APIs. etcd tries to
-ensure the strongest consistency and durability guarantees for a distributed
-system. This specification enumerates the KV API guarantees made by etcd.
+etcd is a consistent and durable key value store.
+The key value store is exposed through the KV and Watch [gRPC Services].
```
ahrtr, line 8: Transaction is a core concept of databases, including etcd. Why do you remove "`with [mini-transaction][txn] support`"?

content/en/docs/v3.6/learning/api_guarantees.md
```diff
@@ -1,75 +1,60 @@
...
+etcd ensures the strongest consistency and durability guarantees for a distributed system.
+This specification enumerates the API guarantees made by etcd.
 
-### APIs to consider
+### RPCs to consider
```
ahrtr, line 12:
```suggestion
### APIs to consider
```

content/en/docs/v3.6/learning/api_guarantees.md
```diff
@@ -1,75 +1,60 @@
...
+* KV Service
+  * [Range](../api/#range)
+  * [Put](../api/#put)
+  * [Delete Range](../api/#delete-range)
+  * [Transaction](../api/#transaction)
+* Watch Service
+  * [Watch](../api/#watch-api)
+* Lease Service
+  * [Lease grant](../api/#obtaining-leases)
+  * [Keep alive](../api/#keep-alives)
```
ahrtr, line 23: We are talking about APIs in this doc, why introduce another concept "service"? 
```suggestion
* KV APIs
  * [Range](../api/#range)
  * [Put](../api/#put)
  * [Delete](../api/#delete-range)
  * [Transaction](../api/#transaction)
* Watch APIs
  * [Watch](../api/#watch-api)
* Lease APIs
  * [Grant](../api/#obtaining-leases)
  * [Revoke]
  * [Keep alive](../api/#keep-alives)
  * [Put (attaching a lease to a key)]
```

ahrtr (changes requested):

content/en/docs/v3.6/learning/api_guarantees.md
serathius: Let's continue the discussion below.

content/en/docs/v3.6/learning/api_guarantees.md
serathius: Because it's not really the "canonical" feature of etcd. It's not something unique to etcd, especially when we compare to watch.

content/en/docs/v3.6/learning/api_guarantees.md
serathius: FYI etcd is not transactional database. It's key value store, that's why we don't invoke full acid here.

content/en/docs/v3.6/learning/api_guarantees.md
serathius: Please read the discussion on the document https://docs.google.com/document/d/1NUZDiJeiIH5vo_FMaTWf0JtrQKCx0kpEaIIuPoj9P6A/edit?disco=AAAAv5-fWt0

Separation of Resumable from Bookmarkable was proposed by @jpbetz. Paraphrasing:
There are many streaming services that do not provide resumable guarantee, so the fact that etcd does is pretty important. For example zookeeper does not provide resumable. https://zookeeper.apache.org/doc/current/zookeeperProgrammers.html#sc_WatchSemantics

content/en/docs/v3.6/learning/api_guarantees.md
```diff
@@ -106,9 +109,35 @@ expired by the wall clock time to live (TTL). However, users need to be aware
...
+### Revision
+
+An etcd operation that modifies the key value store is assigned a single
+increasing revision. A transaction operation might modify the key value store
+multiple times, but only one revision is assigned. The revision attribute of a
+key value pair that was modified by the operation has the same value as the
+revision of the operation. The revision can be used as a logical clock for key
+value store. A key value pair that has a larger revision is modified after a key
+value pair with a smaller revision. Two key value pairs that have the same
+revision are modified by an operation "concurrently".
```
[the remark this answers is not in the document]
serathius: Because important part should be at the top. You cannot ask someone reading the document about api guarantees to go first for very wordy section of definitions first. You give them the important parts first and if they don't know some term, they can use link to jump to their definition.

content/en/docs/v3.6/learning/api_guarantees.md
ahrtr: > FYI etcd is not transactional database. It's key value store, that's why we don't invoke ACID here.

etcd is a K/V store, and it also supports transaction, and supports the four fundamental attributes ACID.

> Please explain what you mean by "original structure".

https://etcd.io/docs/v3.5/learning/api_guarantees/#guarantees-provided

serathius: Those are conditional statements not real transactions as you cannot revert them.  Please note that ACID was never mentioned in the document. If you have suggestion about structure let's discuss them. For me problem with the old document was lack of structure. 

Old document "structure":
* atomicity
* durability
* Isolation level and consistency
  * strict serializability
  * linearizability
  
Problems with it:
* strict serializability implies atomicity and linearizability.
* Isolation level is not a guarantee.

Proposed structure
* Durability
* Strict serializability, which implies
  * Atomicity
   * Linearizability
   
If you think this is not an improvement, please propose some structure. Note I left atomicity and linearizability to provide more granularity when explaining strict serializability. I could remove them if you prefer.

content/en/docs/v3.6/learning/api_guarantees.md
```diff
@@ -85,18 +70,36 @@ most current value. Without linearizability guarantee, the returned value,
...
+## Watch Service
+
+Watches make guarantees about events:
+* Ordered - events are ordered by revision.
+  An event will never appear on a watch if it precedes an event in time that
+  has already been posted.
+* Unique - an event will never appear on a watch twice.
+* Reliable - a sequence of events will never drop any subsequence of events
+  within a compaction window. Watch may be terminated if it's no longer in
+  history window.
```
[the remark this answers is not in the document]
ahrtr: Please 
1. add the "if there are events ordered in time as a < b < c, 
  then if the watch receives events a and c, it is guaranteed to receive b." back, and 
2. define "history window" if you insist on using it.

content/en/docs/v3.6/learning/api_guarantees.md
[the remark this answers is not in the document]
ahrtr: I can also list counter-reasons, but it isn't a big deal, so I won't insist on it.

content/en/docs/v3.6/learning/api_guarantees.md
ahrtr: Indeed Strict serializability implies other items, but the concept of ACID has already been widely accepted. I do think etcd supports ACID although it isn't clearly documented.

I won't insist on this.

content/en/docs/v3.6/learning/api_guarantees.md
ahrtr: I do not see any confusion with regard to "V2" or "V3 API" here. Let's see other maintainer's opinion.

serathius: I didn't mean confusion, I mean that we overabused word "API" because we couldn't use specific term like "method"/"service"/"rpc". For me using term "API" for everything muddies the water and hides the real hierarchy. 

There is one "etcd API", that consists of multiple services like "KV service", which exposes multiple rpcs like "Range" or "Put". This is naming is borrowed from grpc which is used in our api documentation https://etcd.io/docs/v3.5/learning/api/.

I'm not attached to grpc naming, but we should pick up a unified naming scheme. For now being consistent with https://etcd.io/docs/v3.5/learning/api/ seemed like a best option.

content/en/docs/v3.6/learning/api_guarantees.md
serathius: Yea, I noticed a problem with two independent models for naming consistency guarantee. ACID coming from SQL databases, and [consistency models](https://jepsen.io/consistency) coming from distributed systems.

I don't claim to know which is better for etcd documentation, so I wanted to leave the previously present model borrowed from jepsen.

content/en/docs/v3.6/learning/api_guarantees.md
ahrtr: Let's see other maintainers' opinion. cc @mitake @ptabor @spzala

serathius: Linked document was already shared with all [maintainers over two weeks ago](https://groups.google.com/g/etcd-maintainers/c/wuUr5TZaw_E), so I don't think we will get more. I encourage everyone to join the discussion on the document as this represents agreement between etcd and Kubernetes communities.

content/en/docs/v3.6/learning/api_guarantees.md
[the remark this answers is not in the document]
spzala: I am okay in using either `the entire available event history` or `history window` but I think it's good to keep `a < b < c ..` example.

spzala (commented):
Thanks @serathius @ahrtr

content/en/docs/v3.6/learning/api_guarantees.md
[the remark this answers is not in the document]
serathius: Problem is that `a < b < c` is not guaranteed assuming that after `a` there was an compaction, and we resumed watch on `c`.

content/en/docs/v3.6/learning/api_guarantees.md
serathius: cc @jberkus on the naming convention

serathius: Applying the suggestion as this discussion is not relevant to goal of this PR.

content/en/docs/v3.6/learning/api_guarantees.md
spzala: Sorry for any confusion. I was saying that instead of removing the proposed changes we keep it but I was thinking in the same line of write up as @ahrtr suggesting. As I said, I don't have strong opinion. I would vote to keep them in whatever we decide at the end. Thanks!

content/en/docs/v3.6/learning/api_guarantees.md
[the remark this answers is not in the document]
spzala: Ok, may be we modify it as something like `it is guaranteed to receive b given no compaction happened` ? Thanks!

content/en/docs/v3.6/learning/api_guarantees.md
serathius: As mentioned above, there is no implication from {Ordered, Unique, Reliable, Atomic} to {Resumable, Bookmarkable}.

ahrtr: > Linked document was already shared with all maintainers over two weeks ago, so I don't think we will get more.

This PR is changing etcd's doc, I am confused why don't we need to get other maintainers' comments? No matter what other doc is, it doesn't mean that we need to update etcd following exactly the same way.

> As mentioned above, there is no implication from {Ordered, Unique, Reliable, Atomic} to {Resumable, Bookmarkable} so the write up is incorrect.

If a watch somehow is broken, the users just need to kick off a new watch starting from the last revision received in a watch event before the break, then etcd still guarantees `{Ordered, Unique, Reliable, Atomic}`, doesn't it mean what you called "Resumable"? 

FYI.  when there is temporary network disconnection between clientv3 and etcdserver, the etcd clientv3 automatically recovers or resumes the watch. https://github.com/etcd-io/etcd/pull/15248#discussion_r1100631660

The same thing to `Bookmarkable`. Thanks to that etcd guarantees {Ordered, Unique, Reliable, Atomic}, then when a client receives a revision, then all events up to the revision must have been already delivered, as long as there is compaction in between.

Please let me know if I misunderstood your point.

content/en/docs/v3.6/learning/api_guarantees.md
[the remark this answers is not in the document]
serathius: Sounds good.

content/en/docs/v3.6/learning/api_guarantees.md
serathius: > > Linked document was already shared with all maintainers over two weeks ago, so I don't think we will get more.
> 
> This PR is changing etcd's doc, I am confused why don't we need to get other maintainers' comments? No matter what other doc is, it doesn't mean that we need to update etcd following exactly the same way.

This is not what I mean. I said that document was already sent to all maintainers two weeks ago so I would expect they would have comment on the document if they had something to day about the contract itself.

> > As mentioned above, there is no implication from {Ordered, Unique, Reliable, Atomic} to {Resumable, Bookmarkable} so the write up is incorrect.
> 
> If a watch somehow is broken, the users just need to kick off a new watch starting from the last revision received in a watch event before the break, then etcd still guarantees `{Ordered, Unique, Reliable, Atomic}`, doesn't it mean what you called "Resumable"?

What I'm meant is that implication is the other direction. `Resumable` requires `{Ordered, Unique, Reliable, Atomic}`, but fact that database has `{Ordered, Unique, Reliable, Atomic}`, doesn't mean it needs to provide `Resumable`. By fact that etcd has compaction `Resumable` is already limited to available history window. You could technically implement database that never compacted, and database that immediately compacted after a request. So the fact that etcd provides `Resumable` and provide it's limitation is important for users that want to build applications on etcd like Kubernetes.

> FYI. when there is temporary network disconnection between clientv3 and etcdserver, the etcd clientv3 automatically recovers or resumes the watch. [etcd-io/etcd#15248 (comment)](https://github.com/etcd-io/etcd/pull/15248#discussion_r1100631660)

Sure, however etcd client resume has its own limits. Kubernetes implements their own resume mechanism.

> The same thing to `Bookmarkable`. Thanks to that etcd guarantees {Ordered, Unique, Reliable, Atomic}, then when a client receives a revision, then all events up to the revision must have been already delivered, as long as there is NO compaction in between.
> 
> Please let me know if I misunderstood your point.

Bookmarkable is required as progress notification is a special type of watch response. Watch response will include events with their mod revision and the last committed revision. Normally mod revisions are unrelated from committed revision, however in progress notification there are no events and last committed revision needs to be sent after all events with mod revision below it have been delivered.

This property was important in finding https://github.com/etcd-io/etcd/pull/15248

serathius:
Addressed all comments, PTAL @ahrtr @spzala @ptabor

ahrtr:
> Addressed all comments, PTAL @ahrtr @spzala @ptabor

https://github.com/etcd-io/website/pull/676/files#r1185642861 isn't resolved yet.

content/en/docs/v3.6/learning/api_guarantees.md
ahrtr: OK, makes some sense to me.

ahrtr: thx

serathius:
> > Addressed all comments, PTAL @ahrtr @spzala @ptabor
> 
> https://github.com/etcd-io/website/pull/676/files#r1185642861 isn't resolved yet.

Ups, forgot to push.

ahrtr (approved):
LGTM

Thanks @serathius

spzala (approved):
lgtm Thanks @serathius

Issue #697, opened 2023-06-14
Libraries and tools


nicknotfun:
https://github.com/microsoft/etcd3 - A recommended V3 client for Node; has not had an update in 3 years. 

This is especially notable as it appears to have an active bug in its election implementation that permits it to have multiple-leaders: https://github.com/microsoft/etcd3/issues/176

jmhbnz:
Thanks for letting us know about this @nicknotfun.  

Taking a look at the repository I agree there hasn't been any real maintenance in ~3 years and that is a major bug.

I would be open to removing the library from our reference list. We can always reinstate at a later date if it becomes maintained again.  Feel free to open a pull request, we welcome new contributors.

At a broader level I think we need to holistically review the full list of existing libraries to understand which ones are still maintained. We also should establish a policy for what bar libraries need to meet in order to be listed, so we can make fast decisions about these in future.

nicknotfun:
https://github.com/etcd-io/website/pull/698

jmhbnz:
Closing - There has been recent maintenance on the library resolving the multiple leaders elected issue.

PR #698, opened 2023-06-16
Remove references to mixer/etcd3

nicknotfun:
As per https://github.com/etcd-io/website/issues/697, this library is currently broken and unmaintained so shouldn't be recommended for now. It can always be added back later.


jmhbnz (approved):
LGTM - Thanks @nicknotfun.  Agree we should not promote libraries that are no longer maintained and have showstopping bugs (in this case elections resulting in more than 1 leader).

spzala (commented):
Thanks @nicknotfun and @jmhbnz The broken part can happen in any tool. We cannot keep testing all the tools that are mentioned on that page. I see few PRs were merged lately and there are an attempts to make changes with recent open PRs. They might be running short on maintenance side but considering the repo is not from any individuals, I would prefer to ask them and see what they recommend. I think good idea that we open an issue in that repo and wait for sometime and see if we get any reply? Also, cc @connor4312 - any thoughts? Thanks!

jmhbnz:
> The broken part can happen in any tool. We cannot keep testing all the tools that are mentioned on that page.

Agreed - I'm not suggesting we do continual testing, so far we rely on the community to raise issues with us if they think we should no longer promote a particular library or tool on our website and that is exactly what has happened here.

> I see few PRs were merged lately and there are an attempts to make changes with recent open PRs. They might be running short on maintenance side.

Only few prs merged last few years are Microsoft bot adding a boilerplate file and a couple trivial doc updates. No code maintenance for roughly three years.

> but considering the repo is not from any individuals,

Yes repo is under an organisation, but if you look at contributors it is only really worked on by one individual.

>  think good idea that we open an issue in that repo and wait for sometime and see if we get any reply?

https://github.com/microsoft/etcd3/issues/176 Has been open since Feb 10th so ~4 months, with no acknowledgement.

Let's see if @connor4312 has any future plans for this project.  With that said, unless there is fix planned for the showstopping bug of multiple leaders I don't think it is responsible for us to promote the library on our website.

connor4312:
My work life is now at vscode and we don't use etcd for anything, and my personal life has been recently busy enough without additional side projects.

I think the library is still valuable, particularly as the only reasonably complete (*minus any additions in recent years) v3 implementation for Node. An unsoundness issue in its election implementation does not render the entire library as "currently broken," and if anyone feels strongly enough about it to submit a PR I would be more than happy to review it. Of course anyone is also free for fork it under the terms of its MIT license if they would like to take up the mantle; I'd offer to transfer ownership but I don't know what the process for that is like in Microsoft, it's not something I've navigated before.

I can try to prioritize it a bit but no commitment.

jmhbnz:
Thanks for the quick reply @connor4312!  So reading between the lines it sounds as though there is currently no maintenance plans on the microsoft side for this library and you're fairly stretched for time at the moment which is fair enough.

@spzala how about we leave this pr open for a while and revisit in the near future to see if anyone is been able to address the elections resulting in multiple leaders issue, or some new maintainer steps forward.

Also - Don't want to give impression the library no longer has value. From my perspective the concern is the accuracy of what is on the etcd project website and not causing problems for etcd users if they are directed towards a library that is not maintained or has bugs. One option for this library might be marking as public archive so people are clear if it is not maintained. Food for thought, let's see what happens.

aroramrinaal (approved):

ahrtr (approved):

spzala:
@jmhbnz based on reply from @connor4312 I think good to keep the library as a reference, as it's today, but let's keep the PR open but I will go with majority if others don't feel such a need. For all the third party libraries listed, we cannot guarantee they will always work. They are listed as options for different use cases ,and it's usage is something the etcd users can decide. Thanks!

ahrtr:
Sorry, I missed the comment "`I think the library is still valuable`" from @connor4312 , and comment "`I think good to keep the library as a reference`" from @spzala . 

Let me revert the change.

ahrtr:
FYI. https://github.com/etcd-io/website/pull/711

Issue #653, opened 2023-03-17
Add example on how to bootstrap a cluster via v3 discovery


ahrtr:
Add an example into the doc https://etcd.io/docs/v3.6/dev-internal/discovery_protocol/ to help users understand the process

ahrtr:
Linked to https://github.com/etcd-io/website/pull/583

PR #708, opened 2023-07-11
Update clustering.md

david-caro:
Add a note that etcd v2 is needed to run the discovery service.

### DIFF
--- a/content/en/docs/v3.6/op-guide/clustering.md
+++ b/content/en/docs/v3.6/op-guide/clustering.md
@@ -228,7 +228,7 @@ Moreover, discovery URLs should ONLY be used for the initial bootstrapping of a 
 
 #### Custom etcd discovery service
 
-Discovery uses an existing cluster to bootstrap itself. If using a private etcd cluster, create a URL like so:
+Discovery uses an existing cluster to bootstrap itself, note that currently it needs etcd v2 to be supported on that cluster. If using a private etcd cluster, create a URL like so:
 
 ```
 $ curl -X PUT https://myetcd.local/v2/keys/discovery/6c007a14875d53d9bf0ef5a6fc0257c817f0fb83/_config/size -d value=3


content/en/docs/v3.6/op-guide/clustering.md
```diff
@@ -228,7 +228,7 @@ Moreover, discovery URLs should ONLY be used for the initial bootstrapping of a
 
 #### Custom etcd discovery service
 
-Discovery uses an existing cluster to bootstrap itself. If using a private etcd cluster, create a URL like so:
+Discovery uses an existing cluster to bootstrap itself, note that currently it needs etcd v2 to be supported on that cluster. If using a private etcd cluster, create a URL like so:
```
ahrtr, line 231: V2 discovery is deprecated in 3.6. Instead, users are recommended to use v3 discovery.

FYI. https://etcd.io/docs/v3.6/dev-internal/discovery_protocol/

This guide needs to be updated.

jberkus: I don't quite follow.  Given that it's deprecated, why are we recommending that people use it?

ahrtr: Right, both v2 discovery and the public https://discovery.etcd.io/ service are deprecated. Instead, we should recommend users to use v3 discovery https://etcd.io/docs/v3.6/dev-internal/discovery_protocol/. 

I should add an example on how to use v3 discovery to bootstrap new clusters, but unfortunately I do not get time to do it so far. See https://github.com/etcd-io/website/issues/653. It would be appreciated if anyone can help to finish it.

jmhbnz:
Discussed during sig etcd triage meeting.  Thanks for raising this suggestion @david-caro, we will close in favor of https://github.com/etcd-io/website/issues/653.

Happy to re-open this if the original author has time to update for v3 discovery as described in the issue.

PR #756, opened 2023-12-09
Fix the location of the Node.js LTS download link

ivanvc:
This is a minor fix to the README, as I was working on another PR. While setting up my local environment, I noticed that the Node.js LTS link was broken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,6 @@ Follow these steps to add documentation for a newly released version of etcd, vX
 [etcd.io]: https://etcd.io
 [Hugo]: https://gohugo.io
 [localhost:8888]: http://localhost:8888
-[LTS release]: https://nodejs.org/en/about/releases/
+[LTS release]: https://nodejs.org/en/download/
 [Netlify]: https://netlify.com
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -88,6 +88,6 @@ Follow these steps to add documentation for a newly released version of etcd, vX
 [etcd.io]: https://etcd.io
 [Hugo]: https://gohugo.io
 [localhost:8888]: http://localhost:8888
-[LTS release]: https://nodejs.org/en/about/releases/
+[LTS release]: https://nodejs.org/en/download/
```
ivanvc, line 91: Or it could be https://github.com/nodejs/release#release-schedule, which seems the most similar to that old page linked.

jmhbnz (approved):
LGTM - Thanks @ivanvc.  `https://nodejs.org/en/download` seems reasonable to me as it lands on the LTS download by default 🙂

ahrtr (approved):

spzala (approved):
lgtm. I agree with @jmhbnz Download page seems more relevant as it will take you to the LTS release which is what we require. Thanks @ivanvc

Issue #718, opened 2023-08-16
Upgrade Docsy


chalin:
Docsy has had a number of fixes and improvements since it was last updated. By updating we'll fix #498, among other things. As was mentioned in #717, the upgrade is best done in not-too-big version leaps, while examining the generated site page diffs to ensure that nothing has been seriously broken.

nate-double-u:
I took a look at updating this as a Hugo module, but after having chatted with @chalin, I think that's the wrong path. Probably we should move this from being a git submodule to being an npm install.

nate-double-u:
Currently the site pulls docsy in from a CNCF Fork of docsy found here: https://github.com/cncf/docsy/tree/etcd.io

We can update in place there, but should eventually obsolete it by bringing Docsy in via NPM.

nate-double-u:
fixed by: https://github.com/etcd-io/website/pull/803

PR #758, opened 2023-12-17
Introduce dependabot to stay on top of npm ecosystem security

jmhbnz:
Currently security patching for the etcd website npm dependencies relies on a maintainer remembering to review the contents of `package.json` and manually check for updates running something like `npm audit`.

As a result most of the time it just doesn't happen and vulnerabilities go unpatched for long periods of time.

We can automate staying on top of these by adding ` dependabot.yml` so we can more proactively know when updates are required. There are not a large number of dependencies so it should not be too noisy.

Note: I've ignored `hugo-extended` as we know that is blocked until docsy upgrade is done.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: /
+    schedule:
+      interval: weekly
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,7 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    ignore:
+      # We can't bump hugo-extended further than 0.110.0.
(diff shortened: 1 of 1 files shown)


spzala (approved):
Dependabot makes sense, thanks @jmhbnz for thinking this. @nate-double-u @chalin @ahrtr thoughts from you? Thanks!

ahrtr (approved):

jmhbnz:
Discussed during sig-etcd triage meeting. Hey @nate-double-u - Can you please take a look at this small pr and let us know if you are ok with it?

nate-double-u:
I rescind my comment -- the way Docsy is currently setup it won't (or shouldn't) automatically fail this check, and i've updated https://github.com/etcd-io/website/issues/718 to mention that we should move it from git submodule to NPM install, where this check would be helpful.

Issue #378, opened 2021-06-15
Default config in v3.4 is different from code


divanodestiny:
[website](https://github.com/etcd-io/website/blob/main/content/en/docs/v3.4/op-guide/configuration.md#--backend-bbolt-freelist-type) vs [doc](https://github.com/etcd-io/etcd/blob/v3.4.16/Documentation/op-guide/configuration.md#--experimental-backend-bbolt-freelist-type)

jmhbnz:
Reviewing this old issue while doing triage I can confirm this is still an issue, thanks for reporting it @divanodestiny  and sorry this did not get reviewed sooner, the etcd project has not had enough contributors to respond to all issues.

 The [current documentation for v3.4](https://etcd.io/docs/v3.4/op-guide/configuration/#--backend-bbolt-freelist-type) states:

```markdown
### --backend-bbolt-freelist-type
+ The freelist type that etcd backend(bboltdb) uses (array and map are supported types).
+ default: map
+ env variable: ETCD_BACKEND_BBOLT_FREELIST_TYPE
```

The [code on `release-3.4` stable branch](https://github.com/etcd-io/etcd/blob/e7da7ebf7e61a98e01a26746856f8a6a2c896294/etcdmain/help.go#L218-L219) states:

```go
  --experimental-backend-bbolt-freelist-type 'array'
    ExperimentalBackendFreelistType specifies the type of freelist that boltdb backend uses(array and map are supported types).
```

https://github.com/etcd-io/etcd/pull/11262 promoted the feature from experimental for `release-3.5` only but somehow the documentation update changing the flag from experimental also got applied to the 3.4 docs.

This would be a quick fix to just update the flag in the v3.4 op guide to `--experimental-backend-bbolt-freelist-type`  and update the default to `array` so the docs match the v3.4 code.

ahrtr:
Resolved in https://github.com/etcd-io/website/pull/778

PR #778, opened 2024-03-02
correct default config

divanodestiny:
correct default config val of `--backend-bbolt-freelist-type` for v3.4
see  #378


jmhbnz (approved):
LGTM - Many thanks for tidying this up @divanodestiny

ahrtr (approved):
lgtm

Thank you!

Issue #771, opened 2024-02-21
Create a System Overview


dwelsch-esi:
## Overview

Rename the "Learning" section to "System Overview" or "Conceptual Overview". This is the place for detailed explanations of the system philosophy, design, and architeture. Edit the content, limiting it to explaining concepts. Move instructional and reference topics to their own areas of the documentation.

Move this new System Overview section to earlier in ToC, perhaps before "Installation".

Audience: All

Type: Conceptual

The existing documentation might contain helpful source material that you can pull into this doc change. See recommendations for the existing (at the time of the CNCF tech doc analysis) etcd technical documentation pages:
https://github.com/cncf/techdocs/blob/main/assessments/0010-etcd/etcd-implementation.md/#general-reorganization

## 🎤 Context

This issue tracks recommended changes resulting from an analysis of the etcd documentation commissioned by CNCF. The analysis and supporting documents are here: https://github.com/cncf/techdocs/tree/main/assessments/0010-etcd/

## ✌️ Possible Implementation

The system overview will have several subsections. The existing subsections are a good start, but don't be shy about adding, consolidating, removing, and renaming sections as needed to best describe the product architecture and design. 

Related material in the current doc:
- https://etcd.io/docs/v3.5/learning/data_model/
- https://etcd.io/docs/v3.5/learning/design-client/
- https://etcd.io/docs/v3.5/learning/design-learner/
- https://etcd.io/docs/v3.5/learning/design-auth-v3/
- https://etcd.io/docs/v3.5/learning/api/
- https://etcd.io/docs/v3.5/learning/api_guarantees/
- https://etcd.io/docs/v3.5/learning/why/
- https://etcd.io/docs/v3.5/learning/glossary/
- https://etcd.io/docs/v3.5/dev-internal/modules/
- https://etcd.io/docs/v3.5/op-guide/runtime-reconf-design/
- https://etcd.io/docs/v3.5/op-guide/failures/
- https://etcd.io/docs/v3.5/dev-guide/api_grpc_gateway/
- https://etcd.io/docs/v3.5/dev-guide/grpc_naming/

Suggested changes:

Recommendations follow for the existing subsections of "Learning":

**Data model**: Include in architecture overview.

**etcd client design**: Include in architecture overview. Remove the Glossary and merge it into the site-wide Glossary.

**etcd learner design**: Include in architecture overview. Eliminate references to previous and future releases; describe the current release only. Put comparisons to other releases in the release notes, if relevant.

[...]

PR #843, opened 2024-05-07
Reorganized, renamed, and moved the Technical Overview information (p…

dwelsch-esi:
Updated, reorganized,and renamed the technical overview (previously "Learning") per:
https://github.com/etcd-io/website/issues/771#issue-2147186171

Did some minor rewriting. Moved the Glossary out of "Learning" into the main table of contents after "FAQ".


jmhbnz (commented):
Thanks for proposing this @dwelsch-esi.  The first question in my mind was given this is quite a substantial restructure is it something we should apply to all supported docs versions at the same time? I.e. include `v3.4` and `v3.5`?

My concern would be users getting confused moving between docs for say `v3.5` and `v3.6` etc.

Curious to hear what the community thinks on this.

dwelsch-esi:
@jmhbnz Balance that against confusing users who are familiar with the current documentation. My take would be to treat it like a software update: If it's a bug fix, it might be appropriate to also apply it to existing versions, but since it's (essentially) an enhancement, put it in the new release only. There might be considerations I haven't thought of, though. If the project votes to apply it to other versions, I'm happy to do so.

ivanvc:
Thanks for the PR, @dwelsch-esi. I briefly looked into it but felt overwhelmed, as there were too many changes. I would suggest:

1. Break down the PR into more atomic ones; that way, it will be better for them to merge quicker and lessen the burden on the reviewers.
2. I agree with @jmhbnz. It may be confusing if the documents are not there when switching versions. One approach would be to add redirections to the new locations rather than just moving the files.

content/en/docs/v3.6/tech-overview/api_guarantees.md
```diff
@@ -23,55 +22,55 @@ This specification enumerates the API guarantees made by etcd.
...
-Lease API allows assigning a time to live to a key.
+The KV API allows for direct reading and manipulation of key value store.
+
+The Watch API allows subscribing to key value store changes.
+
+The Lease API allows assigning a time to live to a key.
 
-Both KV and Watch APIs allow access to not only the latest versions of keys, but
-also previous versions are accessible within a continuous history window, limited
+Both the KV and Watch APIs allow access to the latest versions of keys, plus all previous versions within a continuous history window limited
```
nate-double-u, line 31: Often times we use short lines (~80 chars or so) to help localization teams to make their updates.

content/en/docs/v3.6/tech-overview/design-client.md
```diff
@@ -138,3 +132,5 @@ Client-side keepalive ping still does not reason about network partitions. Strea
 ![client-balancer-figure-07.png](../img/client-balancer-figure-07.png)
 
 Currently, retry logic is handled manually as an interceptor. This may be simplified via [official gRPC retries](https://github.com/grpc/proposal/blob/master/A6-client-retries.md).
+
+*Authors: Gyuho Lee (github.com/gyuho, Amazon Web Services, Inc.), Joe Betz (github.com/jpbetz, Google Inc.)*
```
nate-double-u, line 136: I wonder if we still need author attribution here? This isn't a blog post, and I think that the commit history would show the file authorship.

content/en/docs/v3.6/tech-overview/design-learner.md
```diff
@@ -128,3 +123,5 @@ Reference
 - Use case: [etcd#3715](https://github.com/etcd-io/etcd/issues/3715)
 - Use case: [etcd#8888](https://github.com/etcd-io/etcd/issues/8888)
 - Use case: [etcd#10114](https://github.com/etcd-io/etcd/issues/10114)
+
+*Authors: Gyuho Lee (github.com/gyuho, Amazon Web Services, Inc.), Joe Betz (github.com/jpbetz, Google Inc.)*
```
nate-double-u, line 127: see comment above.

content/en/docs/v3.6/upgrades/upgrade_3_3.md
```diff
@@ -399,7 +399,7 @@ _, err := kvc.Get(ctx, "a")
 +  if s.Code() == codes.Canceled
 ```
```
nate-double-u, line 400: GitHub is rendering this section weirdly for me -- showing things have been changed above the fold. See image for what I mean. Let's double check that the changes reflect your intention.

Screenshot 2024-05-17 at 2 32 04 PM

nate-double-u (commented):
In the case of a feature update or bug fix, I agree we should  update all the affected versions; however, here we're doing a reorg, and should probably only update next version (3.6). The draw back is maintaining future 3.5 updates that would need to be ported to 3.6 as the writers will need to know where things shifted.

As for this update, I don't think it's too big a PR, there's just a lot of moved files. 

Some comments in line, otherwise LGTM

/lgtm

content/en/docs/v3.6/tech-overview/api_guarantees.md
dwelsch-esi: Restored shorter line lengths in this file.

content/en/docs/v3.6/tech-overview/design-client.md
dwelsch-esi: Removed authorship lines.

content/en/docs/v3.6/tech-overview/design-learner.md
dwelsch-esi: Removed authorship lines.

content/en/docs/v3.6/upgrades/upgrade_3_3.md
dwelsch-esi: The code example starting on line 383 is a file diff illustrating the differences between the 3.2 and 3.3 config files. GitHub is interpreting the diff notation and highlighting the added and removed lines. GitHub is confused, since these removes/adds are not actually changes to the repo file.

nate-double-u: Thanks for confirming.

nate-double-u (approved):
/lgtm

jberkus:
I'll shepherd this in but I need to fix the merge conflicts first.

jmhbnz:
> I'll shepherd this in but I need to fix the merge conflicts first.

I'm not sure we have consensus for this to proceed as is.  I would like to see the concern addressed about compatibility of docs between versions.  Ivan had a great suggestion in https://github.com/etcd-io/website/pull/843#issuecomment-2109192633 of using redirects to ensure links work across versions.

Issue #844, opened 2024-05-08
Add GitHub workflow to keep the releases up to date


ivanvc:
We need to update the latest version manually by updating a commit and then a pull request (in `content/en/docs/{v3.4,v3.5}/_index.md`).

We can automate this and allow a GitHub action to do it for us and just approve a PR.

jmhbnz:
Discussed during sig-etcd triage meeting, the existing workflow is running well, although it would be much nicer to also get the pr raising automatically. We can't do it with GitHub actions due to GitHub enterprise security policies for the K8s GHE org, however we could exploring using the k8s generic autobumper https://docs.prow.k8s.io/docs/components/cli-tools/generic-autobumper for prow.

ivanvc:
Link to #845.

ivanvc:
I was reading Slack's #github-management. It looks like APISnoop was trying to achieve something related. I'm linking to their prow job PR: https://github.com/kubernetes/test-infra/pull/33365, as it may be a good reference for what we need to do.

jmhbnz:
Discussed during sig-etcd triage meeting, the initial efforts are working well and options to add full automation is unlikely at this stage, closing for now. Thanks @ivanvc

PR #845, opened 2024-05-08
github/workflows: introduce action to update version tag

ivanvc:
Add a nightly job that gets the latest released etcd version for supported branches (v3.4, v3.5) and create a commit and a pull request to keep it up to date.

An example of how this action works: https://github.com/ivanvc/etcd-website/pull/2. Note that it is a fictional version I used to test (https://github.com/ivanvc/etcd/releases).

The caveat is that we must have enabled "Allow GitHub actions to create and approve pull requests". I'm not sure if that's enabled for the repository, as I don't have access to that. Refer to: https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization#preventing-github-actions-from-creating-or-approving-pull-requests

### DIFF
--- a/.github/workflows/update-release-version-template.yaml
+++ b/.github/workflows/update-release-version-template.yaml
@@ -1,0 +1,28 @@
+---
+name: Reusable Template to Update Release
+
+on:
+  workflow_call:
+    inputs:
+      release:
+        type: string
+        required: true
+
+jobs:
+  update-release:
+    name: Update Release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
+      - id: latest-release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          echo version="$(gh release list --repo etcd-io/etcd --json 'tagName' -q 'map(select(.tagName | test("^'"$(echo "${{ inputs.release }}." | sed s/\\./\\\./g)"'"))) | .[0].tagName')" >> "$GITHUB_OUTPUT"
+      - env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          make update-release-version LATEST_VERSION="${{ steps.latest-release.outputs.version }}"
--- a/.github/workflows/update-release-version.yaml
+++ b/.github/workflows/update-release-version.yaml
@@ -1,0 +1,19 @@
+---
+name: Update Latest Release Versions Nightly
+permissions:
+  contents: write
+  pull-requests: write
+on:
+  schedule:
+    - cron: '25 9 * * *' # runs every day at 09:25 UTC
+  workflow_dispatch:
+
+jobs:
+  v3_4:
+    uses: ./.github/workflows/update-release-version-template.yaml
+    with:
+      release: v3.4
+  v3_5:
+    uses: ./.github/workflows/update-release-version-template.yaml
+    with:
+      release: v3.5
--- a/scripts/update_release_version.sh
+++ b/scripts/update_release_version.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+# This script updates the version of the release in the release file
+# Usage: ./update_release_version.sh <new_version>
+
+set -eo pipefail
+
+if [ "$#" -ne 1 ]; then
+  echo "Usage: $0 <new_version>"
+  exit 1
+fi
+
+new_version="$1"
+release_minor=$(echo "$new_version" | cut -d. -f1-2)
+index_file=content/en/docs/"$release_minor"/_index.md
+git_remote="${GIT_REMOTE:-origin}"
+branch="release-${release_minor}-update-latest-release-to-${new_version}"
+current_branch=$(git symbolic-ref HEAD --short)
+
+if [ -z "$GITHUB_ACTOR" ]; then
+  git_author="$(git config user.name)"
+  git_email="$(git config user.email)"
+else
+  git_author="$GITHUB_ACTOR"
+  git_email="$GITHUB_ACTOR@users.noreply.github.com"
+fi
+
+# Check for prerequisites to run the script.
+if ! which gh >/dev/null; then
+  echo "gh needs to be installed for this script to work"
+  exit 1
+fi
+if ! gh auth status >/dev/null; then
+  echo "gh needs to be authenticated for this script to work"
+  exit 1
+fi
+if ! grep git_version_tag "$index_file" | grep -v -e "$new_version\$" >/dev/null; then
+  echo "nothing to do; file $index_file is already up to date with $new_version"
+  exit 0
+fi
+if git ls-remote --exit-code "$git_remote" --heads refs/heads/"$branch" >/dev/null; then
+  echo "nothing to do; branch $branch already exists"
+  exit 0
+fi
+
+# Switch to a new branch.
+git checkout -b "$branch"
+trap 'git checkout "$current_branch"' EXIT
+
+# Update the release version in the release file.
+sed -i 's/git_version_tag:\sv\([0-9]\+\.\)\{2\}[0-9]\+/git_version_tag: '"$new_version"'/' "$index_file"
+
+# Commit the changes, push and create a PR.
+git add "$index_file"
+git -c user.name="$git_author" -c user.email="$git_email" commit --file=- <<EOL
+[${release_minor}] Update installation version to latest tag (${new_version})
+
+Signed-off-by: ${git_author} <${git_email}>
+EOL
+git push "$git_remote" "$branch"
+gh pr create --fill --body "Automated update for ${release_minor}: ${new_version}"
--- a/Makefile
+++ b/Makefile
@@ -19,3 +19,11 @@ get-link-checker:
 
 clean:
 	rm -rf $(HTMLTEST_DIR) public/* resources
+
+.PHONY: update-release-version
+update-release-version:
+ifndef LATEST_VERSION
+	@echo "LATEST_VERSION needs to be specified" && exit 1
+else
+	./scripts/update_release_version.sh "$(LATEST_VERSION)"
+endif


.github/workflows/update-release-version-template.yaml
```diff
@@ -0,0 +1,38 @@
...
+          index_file="content/en/docs/$minor/_index.md"
+          latest_version="${{ steps.latest-release.outputs.version }}"
+          branch="release-$minor-update-latest-release-to-$latest_version"
+          if ! grep git_version_tag "$index_file" | grep -v -e "$latest_version"'$'; then exit 0; fi
+          git checkout -b "$branch"
+          sed -i 's/git_version_tag:\sv\([0-9]\+\.\)\{2\}[0-9]\+/git_version_tag: '$latest_version'/' "$index_file"
+          git add "$index_file"
+          git -c user.name="$GITHUB_ACTOR" -c user.email="$GITHUB_ACTOR@users.noreply.github.com" commit -m "[$minor] Update installation version to latest tag ($latest_version)"
+          git push origin "$branch"
+          gh pr create -f -b "Automated update for $minor: $latest_version"
```
jmhbnz, line 38: Nit: Where possible I prefer using long version of these parameter names as it saves time looking up what they are for people who don't know them off by heart.

Suggestion below but it applies in other lines above.

```suggestion
          gh pr create --fill --body "Automated update for $minor: $latest_version"
```

.github/workflows/update-release-version-template.yaml
```diff
@@ -0,0 +1,38 @@
...
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          minor=${{ inputs.release }}
+          index_file="content/en/docs/$minor/_index.md"
+          latest_version="${{ steps.latest-release.outputs.version }}"
+          branch="release-$minor-update-latest-release-to-$latest_version"
+          if ! grep git_version_tag "$index_file" | grep -v -e "$latest_version"'$'; then exit 0; fi
+          git checkout -b "$branch"
+          sed -i 's/git_version_tag:\sv\([0-9]\+\.\)\{2\}[0-9]\+/git_version_tag: '$latest_version'/' "$index_file"
+          git add "$index_file"
```
jmhbnz, line 35: Nit: I think shellcheck suggests encapsulating variables as follows:

Perhaps just put the script standalone in a file and run shellcheck against it to double check.

```suggestion
          git add "${index_file}"
```

jmhbnz (commented):
Thanks for pulling this together @ivanvc - initial thoughts below:

.github/workflows/update-release-version.yaml
```diff
@@ -0,0 +1,19 @@
...
+on:
+  schedule:
+    - cron: '25 9 * * *' # runs every day at 09:25 UTC
+  workflow_dispatch:
+
+jobs:
+  v3_4:
+    uses: ./.github/workflows/update-release-version-template.yaml
+    with:
+      release: v34
```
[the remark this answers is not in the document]
ivanvc: Thanks, good catch 😅

.github/workflows/update-release-version-template.yaml
ivanvc: Hey @jmhbnz, thanks for the suggestion. I wasn't also sure if I wanted to place the script straight in this workflow file, or in a bash script in the repository. However, we don't have any other scripts like this in the repo. Would you recommend creating a new `scripts` directory (I don't know if this will break docsy and how we build the website)? I noticed that we have a `Makefile`, but I feel like this script is also particular to this action and may not even work outside of the workflow context. Let me know your thoughts so we can move forward with this :)

.github/workfows/update-release-version-template.yaml
```diff
@@ -0,0 +1,33 @@
+---
```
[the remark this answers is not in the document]
ivanvc: 🤦

.github/workflows/update-release-version-template.yaml
jmhbnz: Initial reaction is I'm happy for this chunk of bash to stay in the workflow only for now. We can do it in iterations.

With that said - If you wanted to make a follow-up issue for moving it into a Makefile recipe perhaps called `update-etcd-version` or something similar then this makes sense to me. It would be nice to be able to run locally to test the script. It would just depend on `gh auth status` of local user which we may need to check for.

.github/workflows/update-release-version-template.yaml
ivanvc: Done

.github/workflows/update-release-version-template.yaml
ivanvc: I initially moved the script to the `Makefile`. But, later, I moved it to its own script file. It was getting messy as an embedded script in the `Makefile`, and it wasn't possible to `shellcheck` it later. If we have more scripts later, we can add them as well.

jmhbnz (approved):
LGTM - Thanks for implementing this @ivanvc

scripts/update_release_version.sh
```diff
@@ -0,0 +1,56 @@
...
+if ! gh auth status >/dev/null; then
+  echo "gh needs to be authenticated for this script to work"
+  exit 1
+fi
+if ! grep git_version_tag "$index_file" | grep -v -e "$new_version\$" >/dev/null; then
+  echo "nothing to do; file $index_file is already up to date with $new_version"
+  exit 0
+fi
+if git ls-remote --exit-code "$git_remote" --heads refs/heads/"$branch" >/dev/null; then
+  echo "nothing to do; branch $(BRANCH) already exists"
```
ivanvc, line 41: I just realized I left this from when this was in the `Makefile`. Changing `$(BRANCH)` to `$branch`.

jmhbnz:
Gentle ping @spzala - Can you please be second pair of eyes on this before we merge?

spzala:
> Gentle ping @spzala - Can you please be second pair of eyes on this before we merge?

@jmhbnz sorry I was away but taking a look now. I will ping you for a couple of questions. Thanks!

jmhbnz:
Would be great to test this now that `v3.5.14` has been released. @spzala Please let us know if any questions or ok to proceed.

spzala:
> Would be great to test this now that `v3.5.14` has been released. @spzala Please let us know if any questions or ok to proceed.

Yes, I will get back to you today @jmhbnz to touch base on security considerations. Thanks!

.github/workflows/update-release-version-template.yaml
```diff
@@ -0,0 +1,28 @@
+---
+name: Reusable Template to Update Release
+
+on:
```
spzala, line 4: @ivanvc @jmhbnz I am not an expert but wondering shouldn't we use `workflow_dispatch:` here as well? With `write` permission, we need to be careful as much as possible. I am sure you have thought about security but still wanted to double check. Thanks! cc @ahrtr @serathius @wenjiaswe

spzala (commented):
Thanks @ivanvc Please see an inline comment.

.github/workflows/update-release-version-template.yaml
jmhbnz: Agree we should use `contents: write` sparingly as there are security considerations.

We have `workflow_dispatch` on the scheduled job that calls this one in order to support running a workflow on demand https://github.com/etcd-io/website/pull/845/files#diff-709e89ffb8e7b46a9736507dd8f21a544b583f4fbc0a1fb01c009f3b41c9178cR9.

@ivanvc if we are using this action to raise a pull request automatically for the version update can we get away with `pull-requests: write` or do we absolutely need `contents: write` as well?

ivanvc: I want to test this first and will get back to you with the answer. Either way, I'll narrow down the permissions. Good catch :).

As per `workflow_dispatch:`, I'm fine removing it. It's just a convenience to run them on-demand, as @jmhbnz stated. Also, these triggers can only be done by people with write access to the repo (ref: [GitHub docs](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow#configuring-a-workflow-to-run-manually)).

ivanvc: Without the `contents: write` permission, I get a 403 when pushing to the repository :face_with_head_bandage:. Refer to this run without `contents: write`: https://github.com/ivanvc/etcd-website/actions/runs/9310254830/job/25627319064.

ivanvc: I tried some other options without pushing to a branch. However, if I have the local commit, GitHub's CLI still asks to push it to a remote.

An alternative would be to keep this action running in a fork, and it opens a PR targeting this repository. That way, we won't need the action with `contents: write` in this repository.

Please let me know what you think.

spzala: Thanks much @ivanvc !! If we have control on merging a PR, that would be great. Also, if I understood correctly, `workflow_dispatch:` is a good security practice to manually approve the workflow. So, I didn't say to remove it but I was suggesting/checking that we should probably add it in [update-release-version-template.yaml](https://github.com/etcd-io/website/pull/845/files#diff-55aabdf214faf6ad31817739890a15aa1bf9b21f98f313dd670d605207fdca92) as well.

ivanvc: @jmhbnz, do you have any thoughts on:

> An alternative would be to keep this action running in a fork, and it opens a PR targeting this repository. That way, we won't need the action with contents: write in this repository.

It could remove the security issue that the action has write access to the repository. On the other hand, it might not be ideal that the action doesn't live in the etcd-io organization.

jmhbnz: Hey Team - I am happy to forge ahead and merge (provided @spzala agrees) for the following reasons:
- This workflow does not run `on: pull_request` meaning it only runs against code we have already merged, and only on our schedule or if we manually trigger it (requires maintainer to do so).
- We will be manually approving and deciding to merge any pull requests it raises
- Any changes to the behaviour of this workflow will be reviewed by our pull request process (and should be closely scrutinized).

spzala: @jmhbnz +1 sounds great. Thank you and @ivanvc so much!

ivanvc:
:tada: :100: :bow: I'll keep an eye on the execution. It should raise two PRs for v3.4.33 and v3.5.14.

ivanvc:
The action partially worked. It detected the new version(s), pushed the branches:

* https://github.com/etcd-io/website/compare/main...release-v3.4-update-latest-release-to-v3.4.33
* https://github.com/etcd-io/website/compare/main...release-v3.5-update-latest-release-to-v3.5.14

But it failed to create the pull request :sob:. I think the reason is that I need someone with admin rights to the repository has to enable GitHub actions to create and approve pull requests. 

* https://github.com/etcd-io/website/actions/runs/9513887466/job/26224901728
* https://github.com/etcd-io/website/actions/runs/9513887466/job/26224901469

Referenece:
* https://github.blog/changelog/2022-05-03-github-actions-prevent-github-actions-from-creating-and-approving-pull-requests/
* https://github.com/orgs/community/discussions/110415

A possible enhancement would be setting the commit author to `github-actions`, rather than impersonating the `$GITHUB_ACTOR`.

jmhbnz:
> But it failed to create the pull request 😭. I think the reason is that I need someone with admin rights to the repository has to enable GitHub actions to create and approve pull requests.

Checked settings - for some reason this option is greyed out, potentially a Kubernetes GitHub Enterprise org policy:

image

This [post](https://stackoverflow.com/questions/72376229/github-actions-is-not-permitted-to-create-or-approve-pull-requests-createpullre/73155199#73155199) appears to confirm an org setting needs to be changed. We would need to raise a request issue on `kubernetes/org` to discuss enabling this.

>  On top of Aviggiano's reply. In case you have an Enterprise level account, the settings to Allow GitHub Actions to create and approve pull requests have to be set at the Enterprise settings level and not at the Organization level.

ivanvc:
Would you say this is reasonable to ask for? Or would you suggest looking for an alternative?

jmhbnz:
> Would you say this is reasonable to ask for? Or would you suggest looking for an alternative?

Looks like others have asked for similar and the issues just sit there going rotten. It will probably be more effort than it will be worth.  A better long term approach might be an actual github bot/application but again that is a lot of work for such a minor issue.

Suggest we reduce workflow functionality to just create the branches which we know works, then we can manually open the pull requests which is trivial. I've just done that for current round of patch releases.

ivanvc:
That sounds good. I'll update the script so it doesn't try to create the pull request and doesn't end with an error. I think some kind of mention/notification would be good so these branches don't slip through the cracks.

ivanvc:
Link to #844

PR #853, opened 2024-05-19
etcd website: added robustness tests meeting information

thedtripp:
Added information about robustness tests meetings with relevant links.

### DIFF
--- a/content/en/community/_index.md
+++ b/content/en/community/_index.md
@@ -21,9 +21,9 @@ community_resources:
     desc: >
       Join contributors and maintainers [online][], every two weeks.
   - title: >
-      [<i class="fas fa-file-alt"></i>Meeting doc][meeting-doc]
+      [<i class="fas fa-file-alt"></i>Meeting docs][community-meeting-doc]
     desc: >
-      For meeting details, consult the [etcd community meeting][meeting-doc] document.
+      For meeting details, consult the [etcd community meeting][community-meeting-doc] and [robustness tests meeting][robustness-tests-meeting-doc] documents.
   - title: >
       [<i class="fab fa-youtube"></i>YouTube][etcd-youtube]
     desc: >
@@ -60,10 +60,14 @@ questions:
 
 {{% cards "main_channels" %}}
 
-You can also chat with other Etcd users and contributors in the #sig-etcd slack
+You can also chat with other etcd users and contributors in the #sig-etcd slack
 channel on [Kubernetes Slack][].
 
-## Community meetings
+## Meeting schedule
+
+Please refer to [the Kubernetes community](https://github.com/kubernetes/community/blob/master/sig-etcd/README.md#meetings) SIG etcd summary for the latest updates to etcd community meeting schedules.
+
+### Community meetings & Issue triage
 
 etcd contributors and maintainers meet [online][] every week, on **Thursday
 at 11 AM** [Pacific Time][], alternating between community meetings and issue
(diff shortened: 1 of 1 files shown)


content/en/community/_index.md
```diff
@@ -15,15 +15,15 @@ main_channels:
       [<i class="fab fa-github"></i>Github Discussions][GD]
     desc: >
       Ask and find answers to your etcd questions.
-community_resources:
+community_meeting_resources:
```
spzala, line 18: I don't think you should change this section title.

spzala (commented):
@thedtripp thanks for the PR. I have couple of comments that I started adding inline but more think of this update related to meeting, probably a good idea that for all the meetings we just point to this link on the `sig-etcd` page so that we don't need double maintenance and stay latest as per changes on the SIG page, https://github.com/kubernetes/community/blob/master/sig-etcd/README.md#meetings  cc @jmhbnz @ahrtr

content/en/community/_index.md
```diff
@@ -63,7 +63,9 @@ questions:
 You can also chat with other Etcd users and contributors in the #sig-etcd slack
```
jmhbnz, line 63: Nit: etcd should always be lowercase, let's fix this while we are updating the page.

```suggestion
You can also chat with other etcd users and contributors in the #sig-etcd slack
```

jmhbnz:
> @thedtripp thanks for the PR. I have couple of comments that I started adding inline but more think of this update related to meeting, probably a good idea that for all the meetings we just point to this link on the `sig-etcd` page so that we don't need double maintenance and stay latest as per changes on the SIG page, https://github.com/kubernetes/community/blob/master/sig-etcd/README.md#meetings cc @jmhbnz @ahrtr

I don't mind the dual update in this case as we don't change meeting details often. I believe having the additional context / details in the website is a better user experience and is more likely to have folks wanting to join versus clicking through to a markdown file with no explanation of what the meetings are really about.

For a long time now https://etcd.io/community has been our single page with all details of how people can interact with our community, would be nice to keep it this way even if it means a little extra pr to keep it in sync every year or so if something changes.

Happy to be overruled if others don't agree.

spzala:
Keeping meeting description is fine, I was suggesting more on pointing to the link to stay up to date on timing in case if anything change (day/hours/frequency) in future and we forget to update time on this page, as it would be bad if someone try to join and couldn't. I don't have a strong opinion to your suggestion and so I am fine either way.

content/en/community/_index.md
```diff
@@ -60,10 +60,12 @@ questions:
 
 {{% cards "main_channels" %}}
 
-You can also chat with other Etcd users and contributors in the #sig-etcd slack
+You can also chat with other etcd users and contributors in the #sig-etcd slack
 channel on [Kubernetes Slack][].
 
-## Community meetings
+## Meeting schedule
+
```
jmhbnz, line 67: Suggest adding a note here as suggested by @spzala.
```suggestion

Please refer to [the Kubernetes community](https://github.com/kubernetes/community/blob/master/sig-etcd/README.md#meetings) SIG etcd summary for the latest updates to etcd community meeting schedules.

```

jmhbnz (commented):
Adding the link to source of truth seems like a good call, thanks for the clarification @spzala.

thedtripp:
@jmhbnz @spzala I committed the proposed changes, then squashed the commits. Thanks for the feedback.

spzala:
> Adding the link to source of truth seems like a good call, thanks for the clarification @spzala.

Thanks @jmhbnz and @thedtripp !!

spzala (approved):

PR #863, opened 2024-06-20
Bump netlify-cli to resolve cve

jmhbnz:
Refer: https://github.com/advisories/GHSA-grv7-fg5c-xmjg

I was expecting dependabot to raise this automatically. Wondering if it hasn't because it's only a dev dependency, updated dependabot config to explicitly include both production and development dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,6 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    allow:
+        dependency-type: "development"
+        dependency-type: "production"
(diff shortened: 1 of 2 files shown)


spzala (approved):
@jmhbnz thanks much for taking care of the issue and updating to latest netlify-cli !!

spzala:
Hopefully the changes will help find out if it answers your question in the issue description as we would expect :)

ivanvc (approved):
LGTM. Thanks, @jmhbnz. A way to test the dependabot configuration that I've been using is to update the configuration on your main branch from your fork. It may be a hassle to keep the branch up to date, but it works when experimenting with its configuration :sunglasses:

PR #868, opened 2024-06-29
Fix dependabot config to ensure allow is an array

jmhbnz:
My last pr didn't quite get the dependabot config right, `allow` is supposed to be an array, refer: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,5 +6,5 @@ updates:
     schedule:
       interval: weekly
     allow:
-        dependency-type: "development"
-        dependency-type: "production"
+      - dependency-type: "development"
+      - dependency-type: "production"


spzala (approved):
Ah, YAML's tricky sometimes :) Thanks for fixing it @jmhbnz

ivanvc:
It sounds like linting YAMLs would be a good improvement. Running `yamllint` against the old `dependabot.yml` file yields:

```
.github/dependabot.yml
  9:9       error    wrong indentation: expected 6 but found 8  (indentation)
  10:9      error    duplication of key "dependency-type" in mapping  (key-duplicates)
```

PR #879, opened 2024-08-06
Add Markdown diff linter

### LINKED ISSUE #857: Enable Markdown linting
Currently, the repository doesn't have a Markdown linter. This causes PRs sometimes to discuss using Markdown standards and contributes to a drift in discussions. Following what we recently discussed in a PR (https://github.com/etcd-io/website/pull/774#discussion_r1498552258). Enabling linting for the whole repository may require a significant effort. I think a good starting point would be to incrementally run a Markdown linter against the changes from pull requests.

ivanvc:
I had this issue open for a while but couldn't get back to it. ~It adds a GitHub workflow~ (EDIT: we decided to make this a standalone script, so it can be used with prow) that links **changed Markdown files** and fails if the issue is **within the changed lines**.

It uses NodeJS' [markdownlint-cli2](https://github.com/DavidAnson/markdownlint-cli2), which has improvements over markdownlint-cli. I introduced a configuration file, `.markdownlint-cli2.yaml`, which allows lines longer than 80 characters, as this rule would raise issues in every Markdown file we have in the repository.

This should help with reviews, avoiding discussing Markdown formatting.

An example of a run at my fork: https://github.com/ivanvc/etcd-website/pull/6.

Fixes #857.

### DIFF
--- a/scripts/markdown_diff_lint.sh
+++ b/scripts/markdown_diff_lint.sh
@@ -1,0 +1,98 @@
+#!/usr/bin/env bash
+# This script runs markdownlint-cli2 on changed files.
+# Usage: ./markdown_lint.sh
+
+set -eo pipefail
+
+if ! command markdownlint-cli2 dummy.md &>/dev/null; then
+  echo "markdownlint-cli2 needs to be installed."
+  echo "Please refer to https://github.com/DavidAnson/markdownlint-cli2?tab=readme-ov-file#install for installation instructions."
+  exit 1
+fi
+
+COLOR_RED='\033[0;31m'
+COLOR_BOLD='\033[1m'
+COLOR_ORANGE='\033[0;33m'
+COLOR_NONE='\033[0m' # No Color
+
+function log_error {
+  echo -n -e "${COLOR_BOLD}${COLOR_RED}$*${COLOR_NONE}\n"
+}
+
+function log_warning {
+  echo -n -e "${COLOR_ORANGE}$*${COLOR_NONE}\n"
+}
+
+if [ -z "${PULL_BASE_SHA}" ]; then
+  echo "Empty base reference (\$PULL_BASE_SHA), assuming: main"
+  PULL_BASE_SHA=main
+fi
+
+if [ -z "${PULL_PULL_SHA}" ]; then
+  PULL_PULL_SHA="$(git rev-parse HEAD)"
+  echo "Empty pull reference (\$PULL_PULL_SHA), assuming: ${PULL_PULL_SHA}"
+fi
+
+MD_LINT_URL_PREFIX="https://github.com/DavidAnson/markdownlint/blob/main/doc/"
+
+mapfile -t changed_files < <(git diff "${PULL_BASE_SHA}" --name-only)
+declare -A files_with_failures start_ranges end_ranges
+
+for file in "${changed_files[@]}"; do
+  if ! [[ "$file" =~ .md$ ]]; then
+    continue
+  fi
+
+  # Find start and end ranges from changed files.
+  start_ranges=()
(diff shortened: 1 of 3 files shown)


ivanvc:
/unhold
cc. @jmhbnz

.github/workflows/markdown-diff-lint.yaml
```diff
@@ -0,0 +1,24 @@
+---
+name: Lint Markdown file changes
```
jmhbnz, line 2: Sorry if this is a naive question - just want to make sure we have thought about it. Should we implement this with prow or GitHub actions?

jmhbnz (commented):
Many thanks for getting this underway @ivanvc.  Question below about pipeline approach, want to cover that off first before we go to much further.

.github/workflows/markdown-diff-lint.yaml
ivanvc: Good point. I'm fine either way. If we decide to go with prow, there are a couple of things that I'm not sure about:

* Is this repo already active in prow? Can I drop a new configuration file, or must we request it be activated?
* There are some required variables that I don't know the equivalent in the prow environment. I'm sure they must be available, but I'm still learning prow :), namely:
   * BASE_REF: github.event.pull_request.base.ref
   * BASE_CLONE_URL: github.event.pull_request.base.repo.clone_url
   * HEAD_CLONE_URL: github.event.pull_request.head.repo.clone_url

To ensure that we are on the same page (and dumping my thoughts so I don't forget ;), we'll need to use a node base image as release-notes does. Refer to their [release-notes-presubmits.yaml](https://github.com/kubernetes/test-infra/blob/73346eb2d68449fb9fec028c0547022fdfa0356d/config/jobs/kubernetes-sigs/release-notes/release-notes-presubmits.yaml#L9-L15).

jmhbnz: Let's try and go to prow in the first instance. I don't think we have any existing jobs for this repo but I don't see why it wouldn't be possible.  We will need to customise https://github.com/kubernetes/test-infra/blob/master/config/jobs/etcd/etcd-periodics.yaml#L8-L11 for example to refer to `website` rather than `etcd`.

Let's refine this pr to just be the scripting / Makefile / markdownlint config and raise a follow-up in test-infra to attempt adding a prow job.

ivanvc: I made the changes to make the script standalone. Can you PTAL? Thanks

scripts/markdown_diff_lint.sh
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env bash
+# This script runs markdownlint-cli2 on changed files.
+# Usage: ./markdown_lint.sh
+
+set -eo pipefail
+
+if ! command markdownlint-cli2 dummy.md &>/dev/null; then
+  echo "markdownlint-cli2 needs to be installed."
+  echo "Install it by running npm install -g markdownlint-cli2"
```
jmhbnz, line 9: Could we instead just link the the markdown-cli2 install docs: https://github.com/DavidAnson/markdownlint-cli2?tab=readme-ov-file#install as personally I found the `brew install` option much more convenient for my device.

jmhbnz (approved):
LGTM - The `make markdown-diff-lint` recipe works well for me locally.

One minor nit about dependency install below, thanks @ivanvc

scripts/markdown_diff_lint.sh
ivanvc: Makes sense. Updated!

spzala (approved):
Great work @ivanvc Thank you!

PR #880, opened 2024-08-15
fix dash replace on flags docs <=3.4

### LINKED ISSUE #870: Option headers in configuration guide transform `--` into an en-dash
The "Configuration Options" page at https://etcd.io/docs/v3.4/op-guide/configuration/ is quite confusing, as the section headers with the option names use `&ndash;` instead of `--`, making it look like etcd wants a single dash rather than a double dash.

moficodes:
fixes #870

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -52,6 +52,9 @@ markup:
   goldmark:
     renderer:
       unsafe: true
+    extensions:
+      typographer:
+        enDash: '--'
   highlight:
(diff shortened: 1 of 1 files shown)


ivanvc (approved):
LGTM. Thanks, Mofi.

Refer to the live preview with the fix: https://deploy-preview-880--etcd.netlify.app/docs/v3.4/op-guide/configuration/

jmhbnz (approved):
LGTM - Thanks for fixing this up @moficodes, nice idea to tweak the markup config!

PR #900, opened 2024-09-20
add partial for version banner

moficodes:
- add partial for version banner
- add a version called latest in the version list

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -103,6 +103,10 @@ params:
   # This menu appears only if you have at least one [params.versions] set.
   version_menu: Versions
   latest_stable_version: v3.5
+  version:
+
+    deprecation_warning: >-
+      This version of etcd is no longer supported. For the latest version, please see the [latest stable version](/docs/latest/).
   versions:
     - version: v3.6
       url: /docs/v3.6/
--- a/content/en/docs/v3.4/_index.md
+++ b/content/en/docs/v3.4/_index.md
@@ -3,7 +3,7 @@ title: v3.4 docs
 cascade:
   version: &vers v3.4
   git_version_tag: v3.4.34
-  is_deprecated: true
+  is_deprecated: false
   exclude_search: true
 linkTitle: *vers
 simple_list: true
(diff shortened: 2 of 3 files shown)


config.yaml
```diff
@@ -103,10 +103,17 @@ params:
...
   latest_stable_version: v3.5
+  version:
+
+    deprecation_warning: >-
+      This version of etcd is no longer supported. For the latest version, please see the [latest stable version](/docs/latest/).
   versions:
     - version: v3.6
       url: /docs/v3.6/
 
+    - version: latest
```
jmhbnz, line 114: Hey @moficodes thanks for tackling this.  The deploy preview is looking good, however can you explain why we need this new version entry added?

I thought `docs/latest` and `docs/next` where meta versions that we manage as redirects in https://github.com/etcd-io/website/blob/bdb510e276299794734e08ddd96f0cb2686bbc59/layouts/index.redirects#L1-L12

moficodes: A local change I fotgot to remove.
Removed now!

jmhbnz: Hmm one other issue, unfortunately the banner is showing on 3.4 https://deploy-preview-900--etcd.netlify.app/docs/v3.4/

3.4 is definitely still supported so we need to tweak the logic in the partial to address I think

moficodes:
@jmhbnz

jmhbnz:
I am still seeing that v3.4 is showing up in the deploy preview as having the deprecation banner? https://deploy-preview-900--etcd.netlify.app/docs/v3.4/

v3.4 is still supported so should not have the banner, please let me know if I am missing something.

ivanvc:
This change looks good to me. One thing worth noting is that the v3.6 docs now show a warning, too. I think it should be fine, but I just wanted to raise awareness of it.

https://deploy-preview-900--etcd.netlify.app/docs/v3.6/

jmhbnz (approved):
LGTM - Thanks @moficodes

ivanvc (approved):
LGTM. Thanks, Mofi.

PR #899, opened 2024-09-19
 added WG etcd operator community meetings to community section

danish9039:
resolves issue #896

WG etcd operator community meetings to community section

Issue #896, opened 2024-09-17
Add WG etcd operator community meetings to community section


ivanvc:
Add a WG etcd operator section in the community page ([content/en/community/_index.md](https://github.com/etcd-io/website/blob/main/content/en/community/_index.md)).

The meetings are scheduled biweekly on Tuesdays at 11 a.m. PT. The Zoom details are the same as the Robustness and Community/Triage meetings.

jmhbnz:
Completed - Thanks @danish9039 🙏🏻

PR #902, opened 2024-09-26
Update community page

ivanvc:
Follow-up on #899.
Part of #896.

* Standardize section headers and content.
* Use AM Pacific Time, rather than a.m. PT.


jmhbnz (approved):
LGTM - Thanks Ivan!

PR #937, opened 2025-01-04
Fix typo

kevburnsjr:
Changes "Thumbstone" to "Tombstone" as per code const and critically acclaimed 1993 film staring Kurt Russel and Val Kilmer.
https://github.com/etcd-io/etcd/tree/master/server/storage/mvcc/revision.go#L32

[image](https://www.youtube.com/watch?v=OdDjlrGBr38&t=8s)

> If we're going to have a future in this town, it's got to have some law and order.


spzala (commented):
Considering Thumbstone (vs thumb stone) is technically not a dict word, I am OK with the changes.

kevburnsjr:
`Thumbstone` is a non-existent concept. The correct term is [Tombstone](https://en.wikipedia.org/wiki/Tombstone_(data_store)).

The linter is failing because the markdown file makes heavy use of html.
I'm not interested in restructuring all the docs just to submit a typo fix.
Recommend disabling `markdownlint` rule [md033](https://github.com/DavidAnson/markdownlint/blob/main/doc/md033.md) until it passes on main.

jmhbnz (approved):
LGTM - Thanks for fixing this @kevburnsjr

PR #947, opened 2025-01-17
Remove refs to `latest` container tags

### LINKED ISSUE #931: Run etcd clusters inside containers
Documentation page referers to latest tag and it is not recomended in [#15382](https://github.com/etcd-io/etcd/issues/15382):

"If so, I would expect that the default etcd documentation does not prompt to use the latest tag (or state that this is not a good practice)."

"latest" tag can't be used [Docker Image Format v1 and Docker Image manifest version 2](https://github.com/etcd-io/etcd/discussions/17811)

jverce:
# Description
Since we don't support/maintain this tag anymore, we're replacing any references to it with the tag for the specific patch versions.

Fixes #931

### DIFF
--- a/content/en/docs/v2.3/docker_guide.md
+++ b/content/en/docs/v2.3/docker_guide.md
@@ -12,13 +12,17 @@ In order to expose the etcd API to clients outside of the Docker host you'll nee
 export HostIP="192.168.12.50"
 ```
 
-The following `docker run` command will expose the etcd client API over ports 4001 and 2379, and expose the peer port over 2380.
-
-This will run the latest release version of etcd. You can specify version if needed (e.g. `quay.io/coreos/etcd:v2.2.0`).
+The following `docker run` command will expose the etcd client API over ports
+4001 and 2379, and expose the peer port over 2380. Note that etcd discourages
+the usage of implicit container version tags, such as `latest`, so make sure you
+always provide a specific version tag. In the example below, we use the latest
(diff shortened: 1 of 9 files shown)


content/en/docs/v2.3/docker_guide.md
```diff
@@ -12,13 +12,17 @@ In order to expose the etcd API to clients outside of the Docker host you'll nee
 export HostIP="192.168.12.50"
 ```
 
-The following `docker run` command will expose the etcd client API over ports 4001 and 2379, and expose the peer port over 2380.
-
-This will run the latest release version of etcd. You can specify version if needed (e.g. `quay.io/coreos/etcd:v2.2.0`).
+The following `docker run` command will expose the etcd client API over ports
```
jmhbnz, line 15: Docs for unsupported versions (i.e. versions < `v3.4`)  are frozen as we no longer maintain these versions.  Can you please tweak the pr so it only updates v3.4, v3.5 and v3.6? Thanks!

jmhbnz (commented):
Thanks @jverce for the quick tidy up here - just need to trim this down to only the supported versions 🙏🏻

content/en/docs/v2.3/docker_guide.md
jverce: Done!

jmhbnz (approved):
LGTM - Thanks @jverce

PR #954, opened 2025-02-18
Update install.md

LLiuJJ:
Add `core.symlinks=true` config for git clone etcd source code.

### DIFF
--- a/content/en/docs/v3.6/install.md
+++ b/content/en/docs/v3.6/install.md
@@ -39,7 +39,7 @@ source by following these steps:
     repo using the following command.
 
     ```sh
-    $ git clone -b {{< param git_version_tag >}} https://github.com/etcd-io/etcd.git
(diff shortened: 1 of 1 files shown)


content/en/docs/v3.6/install.md
```diff
@@ -39,7 +39,7 @@ source by following these steps:
     repo using the following command.
 
     ```sh
-    $ git clone -b {{< param git_version_tag >}} https://github.com/etcd-io/etcd.git
+    $ git clone --config core.symlinks=true -b {{< param git_version_tag >}} https://github.com/etcd-io/etcd.git
```
jmhbnz, line 42: Thanks for the suggestion @LLiuJJ however I'm unable to reproduce the issue:

```bash
# Check suggested git setting
git config --get core.symlinks

# Empty return value, i.e. it is not set

# Clone and build etcd
git clone https://github.com/etcd-io/etcd.git && cd etcd && make build

...

SUCCESS: etcd_build (GOARCH=amd64)
````

PR #976, opened 2025-03-31
[v3.5] Update installation version to latest tag (v3.5.21)

ivanvc:
Update the latest 3.5 release version to v3.5.21.


spzala (approved):
Thanks @ivanvc

ivanvc:
> I think we need to automatically this or remove it (not sure where it's used).

It's used in several places, i.e.: https://etcd.io/docs/v3.5/install/

We can't automate the second part of the automation because GitHub actions lack write permissions (pushing straight to the repository without a pull request). However, James had an idea of when we would do the latest release. We may be able to automate this fully now.

Issue #903, opened 2024-09-28
add - build script to generate improved documentation and config file for project - etcd, website


webdawg:
Hello All!

I am currently implementing a new etcd stack in archlinux, and in these efforts I am using systemd config based environmental variables.  In doing this the aur package in archlinux is using system.d style of configuration.

There is a /usr/lib/systemd/system/etcd.service.d/10-EnvironmentFile.conf that has a reference to the EnvironmentFile=-/etc/conf.d/etcd for the unit and a nice (but needs updated) ENV_* configuration file in the package that is /etc/conf.d/etcd

Currently I am undertaking a reformat and update of the /etc/conf.d/etcd for my work.  I would like to write, eventually, a script to generate my configuration (/etc/conf.d/etcd) file based on the etcd --help output.  I see there was a removal of ENV_ variables after etcd 3.4, and I can understand why.  Who would want to maintain all that?

I rose this in #sig-etcd slack channel recently (https://kubernetes.slack.com/archives/C3HD8ARJ5/p1727313996375469?thread_ts=1727313558.679039&cid=C3HD8ARJ5), and I am submitting an issue here today to see if anyone thinks it may be wise to create a build script based on the output of etcd --help or similar.  The build script would serve two purposes I think.  The first to give folks a great .conf / ENV_ style configuration for their use, and then also generate pieces or parts of the actual https://etcd.io/docs/v3.5/op-guide/configuration/ page

How would I go about doing this?  I just need the highlevel.  Write script in bla (python?), submit to website repo, target X file, and Y file using these paths.  Or even something like see website build script here and figure it out.

Please feel free to comment with guidance or advice.

Here is some sample work for the  ENV_*/.conf file I am building right now.  I am open for suggestions.  For me it is a new take on the eggdrop config style work seen here:  https://github.com/eggheads/eggdrop/blob/develop/eggdrop.conf

Here is a preview of what I am working on:

```
################################################################################
############################## Background ######################################
################################################################################


# Use this file to pull these env into a systemd service.  For instance:
# /usr/lib/systemd/system/etcd.service.d/10-EnvironmentFile.conf
#
# Contents:
# [Service]
# EnvironmentFile=-/etc/conf.d/etcd

# FYI
# Environment variables: every flag has a corresponding environment variable that has
# the same name but is prefixed withETCD_ and formatted in all caps and snake case. For
# example, --some-flag would be ETCD_SOME_FLAG.




################################################################################
############################## Member flags ####################################
################################################################################


[...]

PR #1015, opened 2025-06-03
Create etcd_dash_dash_help_2to_conf.bash.sh

webdawg:
please be gentle

related issue #903

### DIFF
--- a/scripts/etcd_dash_dash_help_2to_conf.bash.sh
+++ b/scripts/etcd_dash_dash_help_2to_conf.bash.sh
@@ -1,0 +1,276 @@
+#!/bin/bash
+
+# This script takes the output of dash dash help of etcd and outputs a CONFIG_FILE_FORMAT=
+# for example:  bash ./scripts/etcd_dash_dash_help_2to_conf.bash $R/sources/etcd_dash_dash_help.current > $R/configurations/defaults/etc/conf.d/etcd
+# .current is the output of etcd --help - and it transforms that into a .conf style config - nice for systemd environment stuff
+# this is our imperfect script, there are many like it, but this one is ours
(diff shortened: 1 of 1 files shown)


jberkus:
@ivanvc @jmhbnz @spzala defer to you on this one.

ronaldngounou:
@webdawg 

1. To pass the DCO check, you have to signoff your commit off all your commit to always include Signed-off-by: Author Name <authoremail@example.com> in every commit message. 
You can also do this automatically by using the -s flag (i.e., git commit -s).

2. Could you please elaborate on the testing you have done to achive a successful execution of thi script and add it to the PR description ?

jberkus:
Also: this is the version of the script to create a config file.  The one that belongs in /website should be the version that generates the output for configuration.md.  They'd both be pretty similar -- or even one script with switches -- but that's what we need in this location.

jberkus (changes requested):
Changes per prior comment.

jberkus:
@webdawg do you think you'll get back to this change?

webdawg:
So I need to move this to scripts for the project then?  This script will output an optional configuration file into the etcd project.  I would / could create an additional script to take that output, and insert it into configuration.md - it would be easier to have a second script because the conf style config example that is outputted is a closer match to the configuration.md.

How do I move this to a different project or do I start the pull request over? 

Also can someone remind me of how I get this as part of a build trigger?  What about tests?

jberkus:
Yes, you'll need to start a new PR because it'll be against /etcd-io/etcd. 

I'd recommend asking for help on adding it to build etc. on #sig-etcd channel, or the etcd-dev mailing list, or in the Etcd community meeting.

jberkus:
Closing this since we haven't heard.

PR #1022, opened 2025-06-06
Fix the issue of incorrect description of "operations." in the TxnModeWriteWithSharedBuffer function

Jason040906:
I am here https://etcd.io/docs/v3.6/feature-gates/ In the functional gating instructions for Alpha or Beta features, it was found that "InitialCorruptCheck" and "TxnModeWriteWithSharedBuffer" have not been fully modified, and the position of "operations." is also not quite correct. I have made necessary modifications to the document and placed it in the correct location.


jberkus (approved):

jberkus:
Thanks for fix!

PR #1060, opened 2025-09-19
Update typos in 3.5 to 3.6 upgrade guide

ahrtr:
cc @spzala @siyuanfoundation @ivanvc @jmhbnz @jberkus


spzala:
Good catch @ahrtr !! Merging it with two approvals in place.

PR #1066, opened 2025-10-02
Ensure v3.6 install instructions will use latest version

jmhbnz:
Just noticed our install instructions for v3.6 still point users to v3.6.0 ⚠️ 

Let's ensure we get pr's automagically to bump this.


jmhbnz:
Gentle ping @ivanvc

PR #1094, opened 2025-12-15
docs(installation): add mise alternative method installation

jylenhof:


jberkus:
Hey, @jylenhof, thanks for the contribution.  However, we're not going to add documentation on installation specific to a 3rd-party OSS tool without discussing it first.  You are welcome to start a discussion via an issue in this repo, or on SIG-Etcd slack, or even in our community meeting.  But we need that first before we make this kind of addition.

PR #1132, opened 2026-03-14
docs: simplify auto-compaction documentation

### LINKED ISSUE #932: Simplify the document for etcd auto-compaction
etcd v3.3 has already been out of support, so we don't need to talk too much about the old behaviour for 3.3 in https://etcd.io/docs/v3.5/op-guide/maintenance/#auto-compaction. It has no any benefit other than confusing the end users. Instead, we just need to summarize the latest behaviour.

If users still need to look into the historical behavior, they just need to look at 3.4 or older version's document.

kovan:
## Summary
- Replaces the version-by-version historical account (v3.0 through v3.3.3) with a clear description of current auto-compaction behavior
- Documents both `periodic` and `revision` compaction modes with dedicated subsections
- Adds recommended retention values per @ahrtr's guidance in the issue
- Documents the compaction interval rules: every hour for periods >1h, every period for periods <=1h
- Keeps the `--auto-compaction-retention=10h` timeline example (useful) but removes the "previously/now" framing
- Applied to v3.5, v3.6, and v3.7 docs; v3.4 left unchanged to preserve historical context

Verified against `server/etcdserver/api/v3compactor/periodic.go` and `revision.go`.

Fixes #932

### DIFF
--- a/content/en/docs/v3.5/op-guide/maintenance.md
+++ b/content/en/docs/v3.5/op-guide/maintenance.md
@@ -18,7 +18,7 @@ Higher `--snapshot-count` holds more Raft entries in memory until snapshot, thus
 
 Since v3.2, the default value of `--snapshot-count` has [changed from from 10,000 to 100,000](https://github.com/etcd-io/etcd/pull/7160).
 
-In performance-wise, `--snapshot-count` greater than 100,000 may impact the write throughput. Higher number of in-memory objects can slow down [Go GC mark phase `runtime.scanobject`](https://golang.org/src/runtime/mgc.go), and infrequent memory reclamation makes allocation slow. Performance varies depending on the workloads and system environments. However, in general, too frequent compaction affects cluster availabilities and write throughputs. Too infrequent compaction is also harmful placing too much pressure on Go garbage collector. See https://www.slideshare.net/mitakeh/understanding-performance-aspects-of-etcd-and-raft for more research results.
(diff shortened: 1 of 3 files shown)


content/en/docs/v3.5/op-guide/maintenance.md
```diff
@@ -68,9 +70,21 @@ When `--auto-compaction-retention=10h`, etcd first waits 10-hour for the first c
...
+etcd compacts on `"latest revision" - 1000` every 5 minutes. For example, when the latest revision is 30000, it compacts on revision 29000.
```
ronaldngounou: Note that it's 5 minutes by default (in prod this value could be different based on the value of `--auto-compaction-retention`

content/en/docs/v3.7/op-guide/maintenance.md
```diff
@@ -85,9 +99,13 @@ $ etcdctl defrag
 Finished defragmenting etcd member[127.0.0.1:2379]
 ```
 
-**Note that defragmentation to a live member blocks the system from reading and writing data while rebuilding its states**.
+{{% alert title="Note" color="info" %}}
```
ronaldngounou: No need to alert here

content/en/docs/v3.5/op-guide/maintenance.md
kovan: Thanks — reworded to "etcd checks every 5 minutes and compacts on..." to separate the check interval from the compaction action. Applied across all three version docs.

content/en/docs/v3.7/op-guide/maintenance.md
kovan: Reverted all alert shortcodes back to bold text.

jberkus:
Looks good, just a few linter issues to fix:  https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/etcd-io_website/1132/pull-website-lint/2041675759864516608

ronaldngounou (approved):

ronaldngounou:
@kovan Could you please squash your commits? Thanks

Issue #1133, opened 2026-03-15
Ameliorate the generated list of metrics page to be updated


ronaldngounou:
Improve the generated list of metrics page, [URL](https://etcd.io/docs/v3.5/metrics/etcd-metrics-latest.txt)
https://etcd.io/docs/v3.5/metrics/
<img width="681" height="138" alt="Image" src="https://github.com/user-attachments/assets/cad55cc0-c4e6-4d8e-b222-79d40dc75e4d" />

Tasks:
- Improve the rendering
- Automatise this process

christophsj:
Hi! I am new to etcd and looking for ways to contribute. I could work on this :)

My idea would be to use  [Content adapters](https://gohugo.io/content-management/content-adapters/) to generate dynamic pages based on the available txt files. Then link to that page instead of plain txt file. Is that what you had in mind? @ronaldngounou 

The page could look something like this:

<img width="1018" height="373" alt="Image" src="https://github.com/user-attachments/assets/9de18c0a-fd3e-4828-9766-0fb13914b78d" />

ronaldngounou:
Thank you @christophsj for looking into this. Yes, using content adapters is a great idea and good step forward. Feel free to send a PR for this.

Also long-term, the task would be to capture the update when new metrics are added in etcd without having to update the text file. Is that also covered by content adapters?

christophsj:
I've created a PR, appreciate any feedback :)

I'd be glad to continue working on the point you mentioned about not having to update the text file. What is the source of truth for metrics if they're not stored in the .txt, where would we pull the data from?

Thanks

ronaldngounou:
Let's hold on a little bit to gather diverse input from other maintenairs after KubeCon. 

In the meantime, on 
> where would we pull the data from? 

I started a thread on sig-etcd on this topic https://kubernetes.slack.com/archives/C3HD8ARJ5/p1774542863800789. I will look up how kubernetes is [doing](https://kubernetes.io/docs/reference/instrumentation/metrics/) to understand similarities.

AR21SM:
/assign
I’ve been working on the **long-term part** of this for the last couple of days and should have a PR up soon.

I spent some time understanding how Kubernetes handles this without depending on .txt files. The general approach there is to scan the codebase and derive metrics information from source definitions, with stability metadata 
I’m following that direction here so the metrics docs can be generated from source instead of relying on **.txt file**. I’ll share a PR once I’ve cleaned up the implementation.

PR #1139, opened 2026-03-24
docs: generate page for metrics per version  to replace plain .txt file

christophsj:
Create pages for displaying a list of metrics per version, instead of linking to a plain .txt file.

This will make it much easier to read and navigate through the metrics.
The page is based on .txt file content, e.g.:

\# name: "etcd_network_peer_round_trip_time_seconds"
\# description: "Round-Trip-Time histogram between peers"
\# type: "histogram"
etcd_network_peer_round_trip_time_seconds_bucket{To="REMOTE_PEER_NODE_ID",le="+Inf"}

will be parsed into a single row in the table with name, description, type.
Also, Labels within the metric (To="REMOTE_PEER_NODE_ID",le="+Inf" in this example) and one example value will be displayed.

metrics.md was renamed/moved to metrics/_index.md to make breadcrumb navigation work. 

{{< metrics-list >}} is no longer necessary because the links to child pages will be automatically generated below the content.

Use simple_list: true to remove the separator between the heading Generated list of metrics and links to the versions

Issue #1133


jberkus (approved):

jberkus:
Thanks!  And let us know if you want to work on the long-term replacement.

christophsj:
@jberkus definitely,  would be glad to keep working on this

PR #1171, opened 2026-06-03
docs: clarify matching behavior for --peer-cert-allowed-cn and --cert-allowed-hostname flags

### LINKED ISSUE #1062: Clarify what "match" means for `--*-cert-allowed-{cn/hostname}` flags
### What would you like to be added?

https://etcd.io/docs/v3.5/op-guide/security/ says the following about the `--peer-cert-allowed-cn` flag (emphasis mine):

> [v3.3.0](https://github.com/etcd-io/etcd/blob/master/CHANGELOG-3.3.md) adds [etcd --peer-cert-allowed-cn](https://github.com/etcd-io/etcd/pull/8616) flag to support [CN(Common Name)-based auth for inter-peer connections](https://github.com/etcd-io/etcd/issues/8262). Kubernetes TLS bootstrapping involves generating dynamic certificates for etcd members and other system components (e.g. API server, kubelet, etc.). Maintaining different CAs for each component provides tighter access control to etcd cluster but often tedious. When --peer-cert-allowed-cn flag is specified, node can only join with **matching common name** even with shared CAs. For example, each member in 3-node cluster is set up with CSRs (with cfssl) as below:

The example provided below (`m1.etcd.local` matches `etcd.local`, not `m2.etcd.local` seems to suggests something like "is a subdomain of" match, but it's not specified explicitly (at least I didn't found it).

https://etcd.io/docs/v3.5/op-guide/configuration/#security reads to me instead as it should be an exact match.

Could the match function/method used be explicitly called out ?

I presume the same matching mechanisms apply to `--peer-cert-allowed-hostname` (use SAN instead of CN if I'm correct) and `--client-cert-allowed-hostname` 

### Why is this needed?

Easier locking down of an etcd cluster
/area documentation

AshrafAhmed9:
Closes #1062

## Summary

Adds two clarifying sentences to the `--peer-cert-allowed-cn` section of the security documentation (v3.5, v3.6, v3.7) to explain the exact matching semantics:

- `--peer-cert-allowed-cn` performs an **exact string comparison** against the certificate's Common Name (CN) field — no wildcards or prefix matching is supported.
- - `--peer-cert-allowed-hostname` and `--client-cert-allowed-hostname` use Go's `x509.Certificate.VerifyHostname()`, which supports both exact hostnames and wildcard entries (e.g. `*.example.com`).
This distinction was not previously documented and caused confusion for users who expected wildcard support in the CN flag.

## Files Changed

- `content/en/docs/v3.5/op-guide/security.md`
- - `content/en/docs/v3.6/op-guide/security.md`
- - `content/en/docs/v3.7/op-guide/security.md`


jberkus:
OK, this looks fine but man do I hate the historical structuring of this doc.  Gonna open an issue about that.

jberkus (approved):

PR #1178, opened 2026-06-10
Update 3.5, 3.6 and 3.7 docs to cover the new option `--v2-deprecation=write-only-skip-check`

ahrtr:
Link to: 
- 3.5: https://github.com/etcd-io/etcd/pull/21897
- 3.6: https://github.com/etcd-io/etcd/pull/21850
- 3.7: https://github.com/etcd-io/etcd/pull/21848

cc @ivanvc @jberkus @spzala

### DIFF
--- a/content/en/docs/v3.5/op-guide/configuration.md
+++ b/content/en/docs/v3.5/op-guide/configuration.md
@@ -127,10 +127,12 @@ The list of flags provided below may not be up-to-date due to ongoing developmen
   Phase of v2store deprecation. Allows to opt-in for higher compatibility mode.
   Supported values:
     'not-yet'                // Issues a warning if v2store have meaningful content (default in v3.5)
-    'write-only'             // Custom v2 state is not allowed (planned default in v3.6)
-    'write-only-drop-data'   // Custom v2 state will get DELETED !
-    'gone'                   // v2store is not maintained any longer. (planned default in v3.7)
+    'write-only'             // Custom v2 state is not allowed (default in v3.6 and v3.7)
+    'write-only-skip-check'  // Custom v2 state is not supported and, if present, will be ignored (available in v3.5.32+, v3.6.13+, and v3.7.0+). Use this option at your own risk.
+    'write-only-drop-data'   // Custom v2 state will get DELETED ! (planned default in v3.8)
+    'gone'                   // v2store is not maintained any longer.
 ```
+
 ### Security
 
 ```nocode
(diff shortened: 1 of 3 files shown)


content/en/docs/v3.5/op-guide/configuration.md
```diff
@@ -128,6 +128,7 @@ The list of flags provided below may not be up-to-date due to ongoing developmen
   Supported values:
     'not-yet'                // Issues a warning if v2store have meaningful content (default in v3.5)
     'write-only'             // Custom v2 state is not allowed (planned default in v3.6)
```
spzala: Since, the the work is done to make it default in v3.6 (not planned anymore), may be we should remove `planned` and say something like (default in v3.6) or `(default in v3.6 and onward)` ?

content/en/docs/v3.7/op-guide/configuration.md
```diff
@@ -130,7 +130,8 @@ The list of flags provided below may not be up-to-date due to ongoing developmen
   Phase of v2store deprecation. Allows to opt-in for higher compatibility mode.
   Supported values:
     'not-yet'                // Issues a warning if v2store have meaningful content (default in v3.5)
-    'write-only'             // Custom v2 state is not allowed (planned default in v3.6)
+    'write-only'             // Custom v2 state is not allowed (default in v3.6 and v3.7)
```
spzala, line 133: nit, but should we say `(default from v3.6)` so that way we don't need to add newer versions? 
*Also, not related to changes in this PR but same for the line above (and other places) where it says `(default in v3.5)`, may be we want it to change to `(default from v3.5)`?

spzala (commented):
@ahrtr lgtm besides couple of thoughts/small comments. Thanks!!

content/en/docs/v3.5/op-guide/configuration.md
ahrtr: Thanks for the suggestion. Since v3.7.0-rc.0 is already out. Just updated the default options to keep it consistent across v3.5, v3.6 and v3.7.

content/en/docs/v3.7/op-guide/configuration.md
ahrtr: > (default from v3.6)

Different versions may have different default option. Updated to keep it consistent across all versions. same as previous comment https://github.com/etcd-io/website/pull/1178#discussion_r3388905633

Thanks

spzala (approved):
@ahrtr thanks much for quickly addressing my comment and also taking care of additional updates (`gone` etc.)!!

content/en/docs/v3.5/op-guide/configuration.md
spzala: Sounds good. Thanks @ahrtr !!

PR #1179, opened 2026-06-11
docs: update v2 store section of upgrade_3_6.md

silentred:
update `upgrade_3_6.md` per https://github.com/etcd-io/etcd/pull/21889#discussion_r3347268983

### DIFF
--- a/content/en/docs/v3.6/upgrades/upgrade_3_6.md
+++ b/content/en/docs/v3.6/upgrades/upgrade_3_6.md
@@ -24,27 +24,54 @@ Before upgrading to 3.6, make sure that [all of your 3.5 members are updated to 
 #### V2 Store
 
 {{% alert title="Note" color="info" %}}
-If the `--enable-v2` flag is not configured or is set to false, no further action is required.
+If the --enable-v2 flag and the ETCD_ENABLE_V2 environment variable have never been enabled (that is, they have always been unset or set to false) on any member since the cluster was created, it is safe to upgrade to etcd v3.6.
+
+Before upgrading, always ensure that the --enable-v2 flag and the ETCD_ENABLE_V2 environment variable are removed or unset on all etcd members, as they are no longer supported in etcd v3.6.
 {{% /alert %}}
 
-If `--enable-v2` or the environment variable `ETCD_ENABLE_V2="true"` **is** configured, additional steps are required to handle the v2store data:
+If `--enable-v2` or the environment variable `ETCD_ENABLE_V2="true"` is (or has ever been) configured on any member, the v2store must be cleaned up before the cluster can be safely upgraded to v3.6. The general workflow is:
 
-1. If there is data in the v2store that needs to be migrated to the v3store, follow the [v2 migration guide](../../../v3.4/op-guide/v2-migration/) to migrate the data.
+1. Run `etcdutl check v2store` on each member to check whether the v2store contains any non-membership (custom) data. Ensure that you use `etcdutl` of v3.5.32 or later, which extends the check to also inspect the WAL records in addition to the v2 snapshot.
 
-2. Remove the `--enable-v2` flag and the `ETCD_ENABLE_V2="true"` environment variable.
+2. The check reports one of the following outputs. Resolve any issues identified by the check, as described in the linked sections.
 
-3. Run the command `etcdutl check v2store` to verify whether the v2store contains any non-membership (custom) data. If no custom data is present, no further action is required.
+    - `No custom content found in both v2store and WAL records` — the v2store is clean. Skip to step 4.
+    - `detected custom content in v2store` — custom v2 content exists in the v2 snapshot only. Follow the steps in [v2store has custom data](#v2store-has-custom-data).
+    - `detected custom v2 content in WAL records` — custom v2 content exists in the WAL records only. Follow the steps in [WAL has custom v2 data](#wal-has-custom-v2-data).
+    - `detected custom v2 content in both v2store and WAL records` — custom v2 content exists in both locations. Follow the steps in both [v2store has custom data](#v2store-has-custom-data) and [WAL has custom v2 data](#wal-has-custom-v2-data).
+    - Anything else — this outcome normally will not happen; if it does, the check may have produced unexpected output, so stop and investigate manually.
 
-4. If custom data is detected in the v2store, apply the following workaround to remove the legacy data:
+3. Repeat step 1 to verify the v2store is clean. If the check still reports custom content, return to step 2 and resolve the remaining issues.
 
-    - Add the flag `--snapshot-count=1` to each etcd instnace that contains custom data in the v2store.
-    - Restart the etcd instances.
-    - Remove the `--snapshot-count=1` flag from (or restore to its original value, if applicable) from all etcd instances.
-    - Restart the etcd instances again.
+4. Once `etcdutl check v2store` reports "No custom content found in both v2store and WAL records" on every member, remove the `--enable-v2` flag and the `ETCD_ENABLE_V2` environment variable from all etcd members.
 
-5. Run the `etcdutl check v2store` command once more to verify that the v2store no longer contains any non-membership (custom) data. At this point, there should be no custom data remaining in the v2store.
+    **All done! You are safe to upgrade to etcd v3.6 now!**
 
-Once these steps are completed, the v2store is clean, and the upgrade process can proceed.
+##### v2store has custom data
+
+If you care about the v2 custom data, or you need to migrate it to the v3store, follow the [v2 migration guide](../../../v3.4/op-guide/v2-migration/) to migrate the data.
+
+If you do not need the v2 custom data, you can either:
+
+- Manually remove the v2 custom data from the data directory, or
+- Start etcd with the flag `--v2-deprecation=write-only-skip-check` so that the v2store check is skipped on startup. Note that this requires `etcd` of v3.5.32 or later, and you use this option at your own risk.
+
+After resolving, re-run `etcdutl check v2store` to confirm the v2store is clean.
+
+##### WAL has custom v2 data
+
+Custom v2 content in the WAL records is removed automatically when the next v2 snapshot is created. With the default snapshot configuration, a new v2 snapshot is taken after a large number of committed transactions. The exact wait time depends on your cluster's write rate; in busy clusters the next snapshot can be created within a few hours, while in quieter clusters it can take longer.
+
+- **Wait for the next v2 snapshot.** If you can tolerate the wait, keep the cluster running; a new v2 snapshot will eventually be taken and the WAL will be cleaned up automatically.
+
+- **Force a snapshot immediately.** If you do not want to wait, apply the `--snapshot-count=1` workaround on each affected member:
+
+    1. Add the flag `--snapshot-count=1` to each etcd member that reports custom WAL data.
+    2. Restart the etcd members.
+    3. Once the next snapshot is created, remove the `--snapshot-count=1` flag (or restore it to its original value, if applicable).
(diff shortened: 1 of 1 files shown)


ahrtr:
We need to rewrite the doc. From a technical perspective, the overall workflow and key steps would look something like the outline below. As for the wording and phrasing, I'd leave that to AI and native speakers. @jberkus @spzala 

```
func workflow() {
    for {
        call check_v2_custom_data

        if no further action is required when calling check_v2_custom_data {
            remove `--enable-v2` from etcd args and cleanup environment variable `ETCD_ENABLE_V2`
            ALL DONE! You are safe to move to etcd v3.6 now!
            break
        }
    }
}

func check_v2_custom_data() {
    If `--enable-v2` or environment variable `ETCD_ENABLE_V2` has never been configured or has been set to false from the beginning of the cluster creation {
         no further action is required. 
         return
    } 
    
    output = Run `etcdutl check v2store` (ensure to use etcdutl of v3.5.32+)
    
    if output is "No custom content found in both v2store and WAL records" {
         no further action is required. 
         return     
    } else if output is "detected custom content in v2store" {
          call v2store_has_custom_data
    } else if output is "detected custom v2 content in WAL records" {
          call wal_has_custom_v2_data
    } else if output is "detected custom v2 content in both v2store and WAL records" {
          call v2store_has_custom_data
          call wal_has_custom_v2_data
    } else {
          something goes wrong, need human intervention
    }
}

func v2store_has_custom_data(){
    if you care about the v2 custom data or they indeed need to be migrated to v3store {
        follow the v2 migration guide
    } else {
        you can either remove the v2 custom data or set option `--v2-deprecation=write-only-skip-check` (ensure you are on etcd v3.5.32+, also use this option at your own risk)
    } 
}

func wal_has_custom_v2_data() {
    Usually users just need to cleanup `--enable-v2` and env `ETCD_ENABLE_V2`, and keep the cluster running for sometime until a new v2snapshot is automatically created. We need to do some test on a real Kubernetes cluster, and provide a rough estimation how long users might need to wait.

    If users don't want to wait, then apply the `--snapshot-count=1` workaround.
}
```

silentred:
Updated by AI

content/en/docs/v3.6/upgrades/upgrade_3_6.md
```diff
@@ -24,27 +24,54 @@ Before upgrading to 3.6, make sure that [all of your 3.5 members are updated to
 #### V2 Store
 
 {{% alert title="Note" color="info" %}}
-If the `--enable-v2` flag is not configured or is set to false, no further action is required.
+If the `--enable-v2` flag is not configured and the `ETCD_ENABLE_V2` environment variable has never been set (or has been set to `false`) since the cluster was created, no further action is required.
```
ahrtr: If `--enable-v2` or `ETCD_ENABLE_V2` is configured as false, we still need to remove or cleanup them before moving to v3.6.

content/en/docs/v3.6/upgrades/upgrade_3_6.md
```diff
@@ -24,27 +24,54 @@ Before upgrading to 3.6, make sure that [all of your 3.5 members are updated to
...
 
-2. Remove the `--enable-v2` flag and the `ETCD_ENABLE_V2="true"` environment variable.
+2. Interpret the output of the check. The check reports one of the following outcomes:
 
-3. Run the command `etcdutl check v2store` to verify whether the v2store contains any non-membership (custom) data. If no custom data is present, no further action is required.
+    - `No custom content found in both v2store and WAL records` — the v2store is clean. Skip to step 5.
+    - `detected custom content in v2store` — custom v2 content exists in the v2 snapshot only. Follow the steps in [v2store has custom data](#v2store-has-custom-data).
+    - `detected custom v2 content in WAL records` — custom v2 content exists in the WAL records only. Follow the steps in [WAL has custom v2 data](#wal-has-custom-v2-data).
+    - `detected custom v2 content in both v2store and WAL records` — custom v2 content exists in both locations. Follow the steps in both [v2store has custom data](#v2store-has-custom-data) and [WAL has custom v2 data](#wal-has-custom-v2-data).
+    - Anything else — the check failed unexpectedly; stop and investigate manually.
```
ahrtr: pls also mention that normally this case won't happen.

ahrtr:
pls also squash the commits

ahrtr:
The workflow is somehow broken. Probably close & reopen can retrigger the workflows?

silentred:
The workflows have all passed

content/en/docs/v3.6/upgrades/upgrade_3_6.md
```diff
@@ -24,27 +24,54 @@ Before upgrading to 3.6, make sure that [all of your 3.5 members are updated to
 #### V2 Store
 
 {{% alert title="Note" color="info" %}}
-If the `--enable-v2` flag is not configured or is set to false, no further action is required.
+If the `--enable-v2` flag was never set and the `ETCD_ENABLE_V2` environment variable was never set (or has always been set to `false`) on any member since the cluster was created, the v2store is clean. In this case, after running `etcdutl check v2store` to confirm, you can skip the data cleanup steps below and proceed directly to removing the `--enable-v2` flag and the `ETCD_ENABLE_V2` environment variable from all etcd members (these options have been removed in etcd v3.6).
```
ahrtr:
```suggestion
If the --enable-v2 flag and the ETCD_ENABLE_V2 environment variable have never been enabled (that is, they have always been unset or set to false) on any member since the cluster was created, it is safe to upgrade to etcd v3.6. Before upgrading, ensure that the --enable-v2 flag and the ETCD_ENABLE_V2 environment variable have been removed or unset on all etcd members, as they are no longer supported in etcd v3.6.
```

ahrtr (approved):
LGTM. Thanks.

cc @jberkus @spzala

content/en/docs/v3.6/upgrades/upgrade_3_6.md
```diff
@@ -24,27 +24,54 @@ Before upgrading to 3.6, make sure that [all of your 3.5 members are updated to
 #### V2 Store
 
 {{% alert title="Note" color="info" %}}
-If the `--enable-v2` flag is not configured or is set to false, no further action is required.
+If the --enable-v2 flag and the ETCD_ENABLE_V2 environment variable have never been enabled (that is, they have always been unset or set to false) on any member since the cluster was created, it is safe to upgrade to etcd v3.6. Before upgrading, ensure that the --enable-v2 flag and the ETCD_ENABLE_V2 environment variable have been removed or unset on all etcd members, as they are no longer supported in etcd v3.6.
```
spzala: To reduce repeat mentioning of the flag and variable, and add continuity from previous sentence, probably better to rewrite second sentence as `Otherwise, before upgrading, ensure that v2store is cleaned up, as it’s no longer supported in etcd v3.6.`

content/en/docs/v3.6/upgrades/upgrade_3_6.md
```diff
@@ -24,27 +24,54 @@ Before upgrading to 3.6, make sure that [all of your 3.5 members are updated to
...
 {{% /alert %}}
 
-If `--enable-v2` or the environment variable `ETCD_ENABLE_V2="true"` **is** configured, additional steps are required to handle the v2store data:
+If `--enable-v2` or the environment variable `ETCD_ENABLE_V2="true"` **is** (or has ever been) configured on any member, the v2store must be cleaned up before the cluster can be safely upgraded to v3.6. The general workflow is:
```
spzala: I think emphasizing just `is` (**is**) is not needed here.

content/en/docs/v3.6/upgrades/upgrade_3_6.md
```diff
@@ -24,27 +24,54 @@ Before upgrading to 3.6, make sure that [all of your 3.5 members are updated to
...
+    - `No custom content found in both v2store and WAL records` — the v2store is clean. Skip to step 5.
+    - `detected custom content in v2store` — custom v2 content exists in the v2 snapshot only. Follow the steps in [v2store has custom data](#v2store-has-custom-data).
+    - `detected custom v2 content in WAL records` — custom v2 content exists in the WAL records only. Follow the steps in [WAL has custom v2 data](#wal-has-custom-v2-data).
+    - `detected custom v2 content in both v2store and WAL records` — custom v2 content exists in both locations. Follow the steps in both [v2store has custom data](#v2store-has-custom-data) and [WAL has custom v2 data](#wal-has-custom-v2-data).
+    - Anything else — this outcome normally will not happen; if it does, the check may have produced unexpected output, so stop and investigate manually.
 
-4. If custom data is detected in the v2store, apply the following workaround to remove the legacy data:
+3. Resolve the issues identified by the check, as described in the linked sections.
```
spzala: I think this step can be merged with step 2 for a better clarity or flow (since user tend to click the linked section when they see link which happens before step 3) - i.e. re-write step 2 as something like below and remove step 3. `2. The check reports one of the following outputs. Resolve any issues identified by the check, as described in the linked sections.`

content/en/docs/v3.6/upgrades/upgrade_3_6.md
```diff
@@ -24,27 +24,54 @@ Before upgrading to 3.6, make sure that [all of your 3.5 members are updated to
...
-    - Restart the etcd instances.
-    - Remove the `--snapshot-count=1` flag from (or restore to its original value, if applicable) from all etcd instances.
-    - Restart the etcd instances again.
+4. Repeat step 1 to verify the v2store is clean. If the check still reports custom content, return to step 2 and resolve the remaining issues.
 
-5. Run the `etcdutl check v2store` command once more to verify that the v2store no longer contains any non-membership (custom) data. At this point, there should be no custom data remaining in the v2store.
+5. Once `etcdutl check v2store` reports "No custom content found in both v2store and WAL records" on every member, remove the `--enable-v2` flag and the `ETCD_ENABLE_V2` environment variable from all etcd members.
 
-Once these steps are completed, the v2store is clean, and the upgrade process can proceed.
+    **All done! You are safe to move to etcd v3.6 now!**
```
spzala: nit, "upgrade to etcd v3.6" instead of "move" for consistency.

spzala (commented):
@ahrtr and @silentred thanks for the heavy lifting with the PR. Lgtm but couple of thoughts inline.

content/en/docs/v3.6/upgrades/upgrade_3_6.md
ahrtr: Good suggestion! Thanks @spzala

content/en/docs/v3.6/upgrades/upgrade_3_6.md
ahrtr: There isn't any v2 custom data to be cleaned in the case, we only need to ensure the flag and environment to be removed and unset. I agree that it seems duplicated. Also, there is no causal relationship between the first and second sentences, so I think we can separate them.

```
If the --enable-v2 flag and the ETCD_ENABLE_V2 environment variable have never been enabled (that is, they have always been unset or set to false) on any member since the cluster was created, it is safe to upgrade to etcd v3.6.

Before upgrading, always ensure that the --enable-v2 flag and the ETCD_ENABLE_V2 environment variable are removed or unset on all etcd members, as they are no longer supported in etcd v3.6.

```

spzala: Ok, that sounds good, thanks @ahrtr !!

silentred:
Updated. Thanks to both of you. @ahrtr @spzala

ahrtr:
It looks perfect now!

@jberkus do you have comment before merging this PR?

spzala (approved):
Thank you @silentred @ahrtr !!

PR #1181, opened 2026-06-18
Remove all references to play.etcd.io

### LINKED ISSUE #1166: Remove all references to play.etcd.io from docs
We have decided to stop maintaining play.etcd.io.

As such, all references to it should be removed from docs, and the website.   References in old blog posts or documentation before v3.5 should not be removed.  Supporting text around it should be removed.  In some cases, you'll need to refer people to the getting started section of the operating guide instead.

This should be a good search-and-replace task for a newcomer, so have at it!

Ameemsaeed:
Resolves #1166

- Remove Play menu item from config.yaml that linked to http://play.etcd.io/play
- Remove play.etcd.io URL pattern from .htmltest.yml ignore list

As per the decision to stop maintaining play.etcd.io, all active references have been removed from the website. References in old blog posts and documentation before v3.5 were not modified per the issue requirements.

### DIFF
--- a/.htmltest.yml
+++ b/.htmltest.yml
@@ -2,7 +2,6 @@ DirectoryPath: public
 IgnoreDirectoryMissingTrailingSlash: true
 IgnoreAltMissing: true
 IgnoreURLs:
-  - "^http://play.etcd.io.*"
   # Manually verified external links below this point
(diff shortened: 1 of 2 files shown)


jberkus (approved):
Well, these are definitely two places that Play exists.

PR #1186, opened 2026-07-02
docs: add v3.7 to v3.6 downgrade guide

### LINKED ISSUE #1173: Create Downgrade instructions for 3.7
We need an downgrade guide from v3.7 to v3.6., to go here: https://etcd.io/docs/v3.7/downgrades/

It can be based on the 3.6 downgrade guide, without all the special instructions because we don't yet know specific issues around downgrading 3.7, if any. 

Ideally, the writer should also test the downgrade instructions using a test cluster that has recently been upgraded.

@nicknikolakakis you want to tackle this one as well?

nicknikolakakis:
Adds the missing downgrade guide for etcd v3.7 to v3.6 at `content/en/docs/v3.7/downgrades/downgrade_3_7.md`, and links it from the downgrades section index.

Fixes #1173

As requested in the issue, the guide is based on the [v3.6 to v3.5 downgrade guide](https://etcd.io/docs/v3.7/downgrades/downgrade_3_6/), without version-specific special instructions since none are known yet for 3.7.

## Testing

The instructions were tested end to end on a local 3-member cluster that was first rolling-upgraded from v3.6.13 to v3.7.0-rc.0 and then rolling-downgraded back to v3.6.13 following exactly the documented steps. All command output in the guide is real output captured from that run. Verified after the downgrade:

- all members report 3.6.13, cluster version 3.6.0, storage version 3.6.0
- `endpoint health` passes on all members
- data written before the upgrade survived the full cycle
- the leader logs `"the cluster has been downgraded","cluster-version":"3.6.0"`

## Notes on differences from the 3.6 to 3.5 guide

- v3.7 introduces no new flags relative to v3.6 (verified by diffing `etcd -h` between v3.7.0-rc.0 and v3.6.13), so no configuration changes are needed; the flags section reflects that.
- The metrics diff lists the metric families new in v3.7 (`etcd_server_request_duration_seconds`, `etcd_debugging_server_watch_send_loop_*`), verified by diffing `/metrics` of idle single-node instances of each version.
- Unlike v3.5, the v3.6 status endpoint reports downgrade info, so downgraded members keep showing `DOWNGRADE ENABLED` true and their storage version until the downgrade completes; the step 6 note covers this instead of the old caveat.

### DIFF
--- a/content/en/docs/v3.7/downgrades/downgrade_3_7.md
+++ b/content/en/docs/v3.7/downgrades/downgrade_3_7.md
@@ -1,0 +1,274 @@
+---
+title: Downgrade etcd from v3.7 to v3.6
+weight: 6600
+description: Processes, checklists, and notes on downgrading etcd from v3.7 to v3.6
+---
+
+In the general case, downgrading from etcd v3.7 to v3.6 can be a zero-downtime, rolling downgrade:
+
+- one by one, stop the etcd v3.7 processes and replace them with etcd v3.6 processes
+- after enabling the downgrade, new features in v3.7 are no longer available to the cluster
+
+Before [starting a downgrade](#downgrade-procedure), read through the rest of this guide to prepare.
+
+### Downgrade checklists
+
+Highlighted differences between v3.7 and v3.6:
+
+#### Difference in flags
+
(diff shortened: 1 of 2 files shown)


jberkus (approved):

jberkus:
This looks great!  Thanks!

PR #1192, opened 2026-07-14
Add /.well-known/security.txt (RFC 9116)

kobihikri:
## What
Add `/.well-known/security.txt` under `static/` so etcd.io serves one.

## Why
`https://etcd.io/.well-known/security.txt` returns 404. [RFC 9116](https://www.rfc-editor.org/rfc/rfc9116) makes `/.well-known/security.txt` the standard, machine-discoverable pointer to a project's security reporting process. etcd already documents one; this makes it discoverable by the convention. Served verbatim from `static/`, with a future `Expires` per §2.5.5. Signed off per DCO.

I used AI assistance to identify this and draft the file; I verified the live 404 myself.

### DIFF
--- a/static/.well-known/security.txt
+++ b/static/.well-known/security.txt
@@ -1,5 +1,5 @@
 Contact: https://github.com/etcd-io/etcd/security/advisories/new
-Expires: 2027-07-14T00:00:00.000Z
+Expires: 2031-07-14T00:00:00.000Z
 Preferred-Languages: en
 Canonical: https://etcd.io/.well-known/security.txt
 Policy: https://github.com/etcd-io/etcd/blob/main/security/README.md
--- a/static/.well-known/security.txt
+++ b/static/.well-known/security.txt
@@ -1,0 +1,5 @@
+Contact: https://github.com/etcd-io/etcd/security/advisories/new
+Expires: 2027-07-14T00:00:00.000Z
+Preferred-Languages: en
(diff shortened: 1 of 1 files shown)


jberkus:
Thank you!  I'll need to look up this practice, and will approve as soon as I verify.

jberkus:
Is this applicable to us?  The RFC says:

> For web-based services, organizations MUST place the "security.txt" file under the "/.well-known/" path, e.g., https://example.com/.well-known/security.txt as per [[RFC8615](https://www.rfc-editor.org/info/rfc9116/#RFC8615)] of a domain name or IP address.

... but we're not a web service; we're an OSS project that happens to have a website.  Is this RFC for a different kind of entity?

static/.well-known/security.txt
```diff
@@ -0,0 +1,5 @@
+Contact: https://github.com/etcd-io/etcd/security/advisories/new
+Expires: 2027-07-14T00:00:00.000Z
```
jberkus:
```suggestion
Expires: 2031-07-14T00:00:00.000Z
```

Realistically, this file is not going to change more than once every 5 years or so.

jberkus:
OK, verified that kubernetes.io is doing this, so we'll add it as well.

Thanks!

jberkus (approved):
Approved with my update.

PR #1200, opened 2026-07-24
Fix  typo in july-23-patch-release.md

ahrtr:
cc @jberkus @spzala @wendy-ha18


spzala (approved):
lgtm, thanks for catching @ahrtr !!

